PR #6, opened 2018-06-29
Feature/signed build

roesslerj:


.travis.yml
```diff
@@ -8,9 +8,14 @@ cache:
   directories:
     - "${HOME}/.m2/"
 
+before_install:
+- openssl aes-256-cbc -K $encrypted_58a7d4f10478_key -iv $encrypted_58a7d4f10478_iv
```
beatngu13, line 12: Personally, I prefer to use long opts in scripts, because when you look at it in a few weeks, you often don't remember what a particular option does (e.g. `-d` at the end).

beatngu13 (changes requested):
In addition to the comments: I cannot install locally anymore.

```
[INFO] --- maven-gpg-plugin:1.6:sign (sign-artifacts) @ recheck ---
gpg: übersprungen "0x99045A9A": No secret key
gpg: signing failed: No secret key
```

.gitignore
```diff
@@ -2,5 +2,8 @@
 .classpath
 .project
 
+# Artefact from signing jars
+retest-gmbh-gpg.asc
```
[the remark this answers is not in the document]
roesslerj: Good question. This artefact then lies in the workspace. I think for a release it makes sense to ignore it, as this is something gitflow checks...
I'll add it to the PR for retest-model as well.

.travis.yml
roesslerj: This is what was autogenerated by the command `travis encrypt-file retest-gmbh-gpg.asc --add`...

PR #14, opened 2018-07-05
Extend TestCaseFinder

beatngu13:
Now includes JUnit 5 (Jupiter) and TestNg.

### DIFF
--- a/src/main/java/de/retest/recheck/testcase/TestFramework.java
+++ b/src/main/java/de/retest/recheck/testcase/TestFramework.java
@@ -1,0 +1,7 @@
+package de.retest.recheck.testcase;
+
+public interface TestFramework {
+
+	boolean isTestCase( String annotationName );
+
+}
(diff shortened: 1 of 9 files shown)


src/main/java/de/retest/recheck/testcase/TestFramework.java
```diff
@@ -0,0 +1,7 @@
+package de.retest.recheck.testcase;
+
+public interface TestFramework {
+
+	boolean isTestCase( String annotationName );
```
beatngu13, line 5: Is this really a proper name for this method? Because we not just look for test cases, but also for setup and teardown methods such as `@Before`.

roesslerj: Well ... what we really want to do is extract the name of the test case, to use it for the naming of the resulting `*.check` file. `@Before` and others are only indicators, to help with that.

What name would you suggest?

PR #15, opened 2018-07-06
Extend TestCaseFinder

beatngu13:


roesslerj (approved):
Was this tested with JUnit 4, 5 and TestNG?

beatngu13:
Only with JUnit 5 locally. It requires Java 8, hence, I cannot create a test for this. TestNG should work, but then I have to look at least for a Maven Plugin that executes TestNG tests. Want me to do so?

roesslerj:
Nope. We assume it works until someone files a bug... :-D

PR #23, opened 2018-07-20
Add missing annotations

beatngu13:
Although we thought it is sufficient to only use test case annotations, we also have to consider annotations where a recheck implementation might be created or a test result gets persisted.

Afterwards, let's release a hotfix (v0.4.1).


roesslerj (approved):
I don't know if this suffices. If we look at annotations, they are hardly every fully qualified...

PR #32, opened 2018-09-24
Extract String contents for recheck into constants

roesslerj:


src/main/java/de/retest/recheck/FileNamerStrategy.java
```diff
@@ -4,6 +4,10 @@
 
 public interface FileNamerStrategy {
 
+	public static final String RECHECK_FILE_EXTENSION = "recheck";
+	public static final String RECHECK_FOLDER_PROPERTY = "de.retest.RecheckFolder";
```
beatngu13: Isn't this `de.retest.recheckDir`? At least that's how it is defined in `retest-core` (`CheckAccessFileImpl#RECHECK_DIR_PROP`).

src/main/java/de/retest/recheck/MavenConformFileNamerStrategy.java
```diff
@@ -15,13 +15,17 @@ public FileNamer createFileNamer( final String... baseNames ) {
 			@Override
 			public File getFile( final String extension ) {
 				final String baseName = Joiner.on( "/" ).join( baseNames );
-				return new File( "src/test/resources/retest/recheck" + File.separator + baseName + extension );
+				return new File( "src/test/resources/retest/"
+						+ System.getProperty( RECHECK_FOLDER_PROPERTY, RECHECK_FOLDER_DEFAULT ) + File.separator
```
beatngu13: The recheck folder is usually a path relative to the project directory, so this wouldn't work?! See recheck-web example config: https://github.com/retest/recheck-web/blob/master/src/test/resources/retest/retest.properties

beatngu13 (changes requested):

src/main/java/de/retest/recheck/FileNamerStrategy.java
roesslerj: True. But this isn't a breaking change yet... only once we delete the other one.

beatngu13 (approved):

PR #34, opened 2018-09-25
Feature/move to java 8

enesoezel:


pom.xml
```diff
@@ -39,18 +39,6 @@
 		</developer>
 	</developers>
 
-	<contributors>
```
beatngu13: You shouldn't remove them, this part has already been fixed. Simply delete this commit.

beatngu13 (changes requested):
Please delete 442923 and increase the version of `org.slf4j:slf4j-api` to `1.7.25`.

beatngu13 (approved):

PR #53, opened 2018-12-17
Feature/add printers format diffs

diba1013:


beatngu13:
I think it's a decent replacement for `toStringDetailed()`, but IMHO the targeted classes (`ActionReplayResult` etc.) should have at least a minimal `toString()` implementation, so they can be used for logging and so forth.

src/main/java/de/retest/recheck/printer/PrinterValueProvider.java
```diff
@@ -0,0 +1,15 @@
...
+
+import de.retest.ui.DefaultValueFinder;
+import de.retest.ui.descriptors.IdentifyingAttributes;
+
+public interface PrinterValueProvider {
+
+	boolean isDefault( String key, Serializable value );
+
+	static PrinterValueProvider of( DefaultValueFinder finder, IdentifyingAttributes attributes ) {
+		return (key, value) -> finder != null && finder.isDefaultValue( attributes, key, value );
```
beatngu13, line 13: Why should we allow `finder` to be `null` (same for `ElementDifferencePrinter`)?

src/main/java/de/retest/recheck/RecheckImpl.java
```diff
@@ -33,6 +37,9 @@
 
 	private TestReplayResult currentTestResult;
 
+	private final Map<String, DefaultValueFinder> usedFinders = new HashMap<>();
```
beatngu13, line 40: This somewhat implies that an instance of `RecheckImpl` may hold multiple `DefaultValueFinder`s and/or it may change at runtime. Can this be? Don't we have a contract?

beatngu13 (changes requested):

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

beatngu13:
Until merge conflicts are resolved.

src/main/java/de/retest/recheck/RecheckImpl.java
```diff
@@ -90,6 +97,9 @@ public void check( final Object toVerify, final RecheckAdapter adapter, final St
 
 	protected ActionReplayResult createActionReplayResult( final Object toVerify, final RecheckAdapter adapter,
 			final String currentStep ) {
+		final DefaultValueFinder defaultFinder = adapter.getDefaultValueFinder();
+		usedFinders.put( currentStep, defaultFinder );
```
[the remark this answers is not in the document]
diba1013: If you want to prevent overwriting of the value. But if the API is used correctly, there should not be multiple calls of this method with the same `currentStep`.

src/main/java/de/retest/recheck/RecheckImpl.java
diba1013: We are collecting all `DefaultValueFinder`s of each `check` to use them later in `capTest` to provide detailed information about the differences with respect to the corresponding default values. Thus it will change during a test method.

src/main/java/de/retest/recheck/printer/PrinterValueProvider.java
diba1013: The `DefaultValueFinder` is passed through the printers and retrieved from [RecheckImpl#usedFinders](https://github.com/retest/recheck/pull/53/files/a65594d5e454d0bbb2c9cba987fc8849bfa0bf2c#diff-8e60368308fb84436b2ad86ded4b895fR40). If someone does not pass an instance back and instead returns `null` in the `RecheckAdapter#getDefaultValueFinder` or for some reason the `currentStep` is not present in `RecheckImpl#usedFinders`, we get `null` for finder.
We could say for the `RecheckAdapter` that it should not return `null`. But for the second part, we need to adjust [RecheckImpl#printer](https://github.com/retest/recheck/pull/53/files/a65594d5e454d0bbb2c9cba987fc8849bfa0bf2c#diff-8e60368308fb84436b2ad86ded4b895fR41) to use a default value.

beatngu13: If we look at the usage of `RecheckAdapter#getDefaultValueFinder()`, we can see that it is used, e.g., within `AttributesDifferenceFinder` without a `null` check. I think we can assume that it is part of the `RecheckAdapter` contract that the corresponding method must not return `null`.

How would you suggest to adapt second part?

src/main/java/de/retest/recheck/RecheckImpl.java
beatngu13: Well, `RecheckImpl` gets the finder via `RecheckAdapter#getDefaultValueFinder()`, which is not context-sensitive. Can't we assume that the type, respectively, behavior of the finder is always the same, but only the instance may changes? I just think the implementation adds more complexity than needed at the moment.

src/main/java/de/retest/recheck/printer/leaf/MultiLeafDifferencePrinter.java
```diff
@@ -0,0 +1,55 @@
...
+
+	@Override
+	public String toString( final LeafDifference difference, final String indent ) {
+		final Printer<LeafDifference> printer = findPrinter( difference );
+		return printer.toString( difference, indent );
+	}
+
+	@SuppressWarnings( "unchecked" )
+	private Printer<LeafDifference> findPrinter( final LeafDifference difference ) {
+		return printers.entrySet().stream() //
```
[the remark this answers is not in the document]
beatngu13: What do you think about 7602c8b09e? We don't need `instanceof` since the type hierarchy for `LeafDifferance` is flat.

PR #54, opened 2018-12-19
Update assertion

enesoezel:
print gui advice only when recheck files exist

Without recheck-files:
image

With recheck-files and diffs:
image

### DIFF
--- a/src/main/java/de/retest/recheck/RecheckImpl.java
+++ b/src/main/java/de/retest/recheck/RecheckImpl.java
@@ -127,17 +127,9 @@ public class RecheckImpl implements Recheck, SutStateLoader {
 		currentTestResult = null;
 		final Set<Object> uniqueDifferences = finishedTestResult.getUniqueDifferences();
 		if ( !uniqueDifferences.isEmpty() ) {
-			final String message;
-			if ( finishedTestResult.hasNoRecheckFiles() ) {
-				message = "'" + suiteName + "'"
-						+ finishedTestResult.toString().substring( finishedTestResult.toString().lastIndexOf( ':' ) );
-			} else {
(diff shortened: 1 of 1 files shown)


src/main/java/de/retest/recheck/RecheckImpl.java
```diff
@@ -127,11 +127,18 @@ public void capTest() {
 		currentTestResult = null;
 		final Set<Object> uniqueDifferences = finishedTestResult.getUniqueDifferences();
 		if ( !uniqueDifferences.isEmpty() ) {
-			throw new AssertionError( "\nA detailed report will be created at \'" + getResultFile() + "\'. " //
-					+ "You can review the details by using our GUI from https://retest.de/review/." //
-					+ "\n\nThe following differences have been found in \'" + suiteName //
-					+ "\'(with " + finishedTestResult.getActionReplayResults().size() + " check(s)):" //
-					+ "\n" + finishedTestResult.toString() );
+			final String message;
```
beatngu13: How about two `private` methods, e.g., `getNoRecheckFilesErrorMessage()` and `getDifferencesErrorMessage()`. Then we could do something like:

```java
String msg = finishedTestResult.hasNoRecheckFiles() ? getNoRecheckFilesErrorMessage() : getDifferencesErrorMessage();
throw new AssertionError( msg );
```

beatngu13 (changes requested):

src/main/java/de/retest/recheck/RecheckImpl.java
enesoezel: Good idea.

beatngu13 (approved):

PR #55, opened 2018-12-19
Feature/clean pom

enesoezel:


pom.xml
```diff
@@ -97,18 +97,6 @@
 			<version>3.2.0</version>
 		</dependency>
 
-		<dependency>
```
beatngu13: AFAIK these dependencies are being used, right? Generally, I think we should declare dependencies explicitly and not rely on transitivity if the corresponding project uses them directly. The version chaos will go away as soon as we have a proper parent POM.

beatngu13 (changes requested):

beatngu13 (approved):

PR #64, opened 2019-01-25
Fix sonarcloud bugs of sut-api code

enesoezel:


src/main/java/de/retest/configuration/Configuration.java
```diff
@@ -22,13 +22,9 @@ static boolean isLoaded() {
 		return instance != null && instance.retestWorkspace != null;
 	}
 
-	public static void ensureLoaded() {
+	public static synchronized void ensureLoaded() {
```
diba1013: I would rather prefer using a special lock with synchronize(lock) since using synchronized in a method locks on the object's instance. As a consequence if a other method uses the another synchronize on a method, which is completely unrelated to this method, we would synchronize without needing to.

src/main/java/de/retest/persistence/migration/XmlMigrator.java
```diff
@@ -47,7 +47,11 @@ public static NamedBufferedInputStream tryToMigrate( final XmlVersionCheckResult
 			for ( int i = 0; i < transformers.size(); i++ ) {
 				final boolean notFinalInputStream = i < transformers.size() - 1;
 				final InputStream previousInputStream = finalInputStream;
-				finalInputStream = transformers.get( i ).transform( finalInputStream );
+				try {
+					finalInputStream = transformers.get( i ).transform( finalInputStream );
+				} catch ( final IOException e ) {
+					throw new RuntimeException( e );
```
diba1013: In the below try-catch we just log the error. Which method is preferred since we exit the execution here? In either case, I would log the error.

src/main/java/de/retest/persistence/migration/XmlTransformer.java
```diff
@@ -57,19 +55,13 @@ public final InputStream transform( final InputStream inputStream ) {
...
 			eventReader.close();
 			eventWriter.flush();
 			eventWriter.close();
-			out.close();
 
 			return new LZ4BlockInputStream( new DeleteOnCloseFileInputStream( tmpFile ) );
-		} catch ( final XMLStreamException e ) {
-			throw new RuntimeException( e );
-		} catch ( final FactoryConfigurationError e ) {
+		} catch ( final XMLStreamException | FactoryConfigurationError | IOException e ) {
```
diba1013: Why do you catch IOException, if you specify in the method that it is allowed to throw?

src/main/java/de/retest/persistence/migration/XmlTransformer.java
```diff
@@ -57,19 +55,13 @@ public final InputStream transform( final InputStream inputStream ) {
 				final XMLEvent nextEvent = eventReader.nextEvent();
 				convert( nextEvent, eventWriter );
 			}
-
 			eventReader.close();
 			eventWriter.flush();
 			eventWriter.close();
```
diba1013, line 71: It has nothing really to do with your PR, but could we use try-with-resources for the above close?

src/main/java/de/retest/ui/diff/IdentifyingAttributesDifferenceFinder.java
```diff
@@ -24,12 +25,8 @@ public IdentifyingAttributesDifferenceFinder( final RecheckIgnore recheckIgnore
...
 	public IdentifyingAttributesDifference differenceFor( final IdentifyingAttributes expected,
 			final IdentifyingAttributes actual ) {
-		if ( expected == null ) {
-			throw new NullPointerException( "Expected cannot be null!" );
-		}
-		if ( actual == null ) {
-			throw new NullPointerException( "Actual cannot be null!" );
-		}
+		Objects.requireNonNull( expected );
+		Objects.requireNonNull( actual );
```
diba1013: We do not need checks here, since the variables are used below. Let the NPE be implicit.

src/main/java/de/retest/ui/diff/IdentifyingAttributesDifferenceFinder.java
```diff
@@ -67,8 +64,14 @@ private static boolean pathDiffers( final IdentifyingAttributes expected, final
 
 	// We treat null == "". This is OK for visible attributeDifferences but less suitable for invisible ones.
 	private static boolean differs( final Object expected, final Object actual ) {
-		return ObjectUtil.isNullOrEmptyString( expected ) ? !ObjectUtil.isNullOrEmptyString( actual )
-				: !expected.equals( actual );
+		if ( ObjectUtil.isNullOrEmptyString( expected ) ) {
+			return !ObjectUtil.isNullOrEmptyString( actual );
+		} else {
```
diba1013: This else can be removed.

src/main/java/de/retest/ui/diff/IdentifyingAttributesDifferenceFinder.java
```diff
@@ -67,8 +64,14 @@ private static boolean pathDiffers( final IdentifyingAttributes expected, final
...
+			if ( Objects.isNull( actual ) ) {
+				return false;
+			}
+			return !expected.equals( actual );
+		}
```
diba1013: I would prefer to use boolean expressions chaining

src/main/java/de/retest/ui/diff/IdentifyingAttributesDifferenceFinder.java
```diff
@@ -67,8 +64,14 @@ private static boolean pathDiffers( final IdentifyingAttributes expected, final
 
 	// We treat null == "". This is OK for visible attributeDifferences but less suitable for invisible ones.
 	private static boolean differs( final Object expected, final Object actual ) {
-		return ObjectUtil.isNullOrEmptyString( expected ) ? !ObjectUtil.isNullOrEmptyString( actual )
-				: !expected.equals( actual );
+		if ( ObjectUtil.isNullOrEmptyString( expected ) ) {
+			return !ObjectUtil.isNullOrEmptyString( actual );
+		} else {
+			if ( Objects.isNull( actual ) ) {
```
diba1013: Did you fix a bug here since this is not in the old code?

src/main/java/de/retest/ui/diff/RootElementDifferenceFinder.java
```diff
@@ -78,7 +78,7 @@ private RootElement findAndRemove( final RootElement expected, final List<RootEl
 	}
 
 	private double compareContainedComponents( final RootElement expected, final RootElement bestWindow ) {
-		return getNumberOfChildDifferences( expected, bestWindow )
+		return (double) (getNumberOfChildDifferences( expected, bestWindow ))
 				/ (getNumberOfChildComponents( expected, bestWindow ) + 1);
```
diba1013: You could just juse  `+ 1.0`
```suggestion 
return getNumberOfChildDifferences( expected, bestWindow ) / (getNumberOfChildComponents( expected, bestWindow ) + 1.0);
```

src/main/java/de/retest/util/ReflectionUtilities.java
```diff
@@ -22,7 +22,7 @@
 
 	public static List<Field> getAllFields( Class<?> clazz ) {
 		final List<Field> result = new ArrayList<>();
-		while ( !clazz.getName().equals( "java.lang.Object" ) ) {
+		while ( clazz.isAssignableFrom( Object.class ) ) {
```
diba1013: isAssignableFrom checks for Instance and superclass, thus you would also take into account `Object`

diba1013 (dismissed):

beatngu13 (changes requested):
The changes are sometimes hard to follow, they should have been separated into multiple micro commits with explaining commit messages.

src/main/java/de/retest/persistence/migration/XmlTransformer.java
```diff
@@ -39,37 +38,39 @@
 
 	public final InputStream transform( final InputStream inputStream ) {
 		try {
-			reset();
```
diba1013, line 42: Must the reset be on top?

beatngu13 (approved):

PR #74, opened 2019-01-31
Feature/fix file format

enesoezel:


src/main/java/de/retest/Properties.java
```diff
@@ -273,14 +273,6 @@ public static File getDefaultSutDir() {
 		return new File( getReTestInstallDir().getParentFile(), "system-under-test" );
 	}
 
-	public static FileOutputFormat getFileOutputFormat() {
-		final String format = System.getProperty( FILE_OUTPUT_FORMAT, FileOutputFormat.ZIP.toString() );
```
beatngu13, line 277: If we remove this, I assume we can also remove the property behind `FILE_OUTPUT_FORMAT`?

beatngu13 (changes requested):

beatngu13 (approved):

PR #82, opened 2019-02-08
Fix value should be given formatted

enesoezel:


src/main/java/de/retest/recheck/ui/descriptors/TextAttribute.java
```diff
@@ -19,7 +19,7 @@ public TextAttribute( final String key, final String value ) {
 	}
 
 	public TextAttribute( final String key, final String value, final String variableName ) {
-		super( key, value != null ? value.trim() : null, variableName );
+		super( key, value, variableName );
```
beatngu13, line 22: How about a unit test that check that the ctor doesn't trim anymore?

beatngu13 (changes requested):

martin-v (approved):

PR #84, opened 2019-02-11
Update deploy provider

modulo11:


.travis.yml
```diff
@@ -22,8 +22,14 @@ script: ci/script.sh
...
 deploy:
-  # deploy artifacts to Maven Central
   - provider: script
     script: ci/deploy.sh
     on:
       tags: true
+  - provider: releases
+    api_key: "${GH_TOKEN}"
+    file_glob: true
+    file: "./target/recheck-*.jar.*"
```
beatngu13: No need to upload the JAR or anything else, we just only want a release with: `draft: true`.

beatngu13: Just like this (we can probably also remove `skip_cleanup: true`): https://github.com/retest/recheck/commit/eaa1ec88538a7edcc708fabd202c1e6480be3e61

beatngu13 (changes requested):

.travis.yml
```diff
@@ -22,8 +22,13 @@ script: ci/script.sh
...
 
 deploy:
-  # deploy artifacts to Maven Central
   - provider: script
     script: ci/deploy.sh
     on:
       tags: true
+  - provider: releases
+    api_key: "${GH_TOKEN}"
+    file_glob: true
```
beatngu13: I assume we can remove this too?

beatngu13 (changes requested):

beatngu13 (approved):

PR #101, opened 2019-02-20
Feature/fix project root finder

modulo11:


src/test/java/de/retest/recheck/configuration/ProjectConfigurationTest.java
```diff
@@ -15,7 +15,9 @@
 	void local_config_file_should_be_created() throws Exception {
 		cut.ensureProjectConfigurationInitialized();
 
-		final Path configFolder = getProjectRoot().resolve( ProjectConfiguration.RETEST_PROJECT_CONFIG_FOLDER );
+		final Path configFolder = getProjectRoot() //
```
beatngu13: It's a test, why not just call `get()`? (In other tests as well.)

beatngu13 (changes requested):

beatngu13 (approved):

PR #107, opened 2019-02-21
Feature/print only unignored diffs

roesslerj:


modulo11:
Tests are red. I'll start reviewing once they are green.

src/main/java/de/retest/recheck/report/ActionReplayResult.java
```diff
@@ -186,7 +185,10 @@ public String getStateFilePath() {
 		return stateFilePath;
 	}
 
-	public List<ElementDifference> getElementDifferences() {
+	/**
+	 * Return <em>all</em> child differences recursively (also child-child differences).
+	 */
+	public List<ElementDifference> getElementAllDifferences() {
```
modulo11: Maybe something like `getAllElementDifferenes` or `getDeepElementDifferences`would be more suitable.

src/main/java/de/retest/recheck/ui/diff/ElementDifference.java
```diff
@@ -85,15 +85,23 @@ public Screenshot mark( final Screenshot screenshot ) {
 		return differences;
 	}
 
-	public List<AttributeDifference> getAttributeDifferences( final ShouldIgnore ignore ) {
+	public List<AttributeDifference> getAttributeDifferences( final ShouldIgnore input ) {
 		final List<AttributeDifference> differences = new ArrayList<>();
+		final ShouldIgnore ignore = input != null ? input : ShouldIgnore.IGNORE_NOTHING;
```
modulo11: Is it possible to just return `differences`If `input` is `null`?

`if ( ignore != null && ignore.shouldIgnoreElement( element ) ) {...}`

src/main/java/de/retest/recheck/ui/diff/ElementDifference.java
```diff
@@ -85,23 +85,27 @@ public Screenshot mark( final Screenshot screenshot ) {
...
 		}
 		if ( attributesDifference != null ) {
-			differences.addAll( attributesDifference.getDifferences().stream()
-					.filter( d -> !ignore.shouldIgnoreAttributeDifference( element, d ) )
-					.collect( Collectors.toList() ) );
+			if ( ignore == null ) {
+				differences.addAll( attributesDifference.getDifferences() );
+			} else {
+				differences.addAll( attributesDifference.getDifferences().stream()
+						.filter( d -> ignore == null || !ignore.shouldIgnoreAttributeDifference( element, d ) )
```
modulo11: It seems that `ignore` can't be null here.

modulo11 (changes requested):

modulo11 (approved):

PR #114, opened 2019-02-24
Replace counter with listener

roesslerj:


modulo11:
Will be discussed when you are back.

PR #118, opened 2019-02-25
Feature/improve root finder

modulo11:


src/main/java/de/retest/recheck/configuration/ProjectConfigurationUtil.java
```diff
@@ -7,15 +7,23 @@
 
 	private ProjectConfigurationUtil() {}
 
-	public static Optional<Path> findProjectConfiguration() {
-		return ProjectRootFinderUtil.getProjectRoot().map( ProjectConfigurationUtil::resolveProperties );
+	public static Optional<Path> findProjectConfigurationFolder() {
+		return ProjectRootFinderUtil.getProjectRoot()
+				.map( path -> path.resolve( ProjectConfiguration.RETEST_PROJECT_CONFIG_FOLDER ) );
```
beatngu13, line 12: If it's the same map function in each method, we could extract it like so:

```java
private Function<Path, Optional<Path>> resolveProjectConfigFolder() {
    return path -> path.resolve( ProjectConfiguration.RETEST_PROJECT_CONFIG_FOLDER );
}

public static Optional<Path> findProjectConfigurationFolder() {
    return ProjectRootFinderUtil.getProjectRoot().map( resolveProjectConfigFolder() );
}

**Review by beatngu13 — CHANGES_REQUESTED** [review]

File: src/main/java/de/retest/recheck/configuration/MavenProjectRootFinder.java [hunk 2]
```diff
@@ -20,22 +20,24 @@ public MavenProjectRootFinder() {
 	}
 
 	@Override
-	public Optional<Path> findProjectRoot( final Path basePath ) {
-		if ( basePath == null || !basePath.toFile().exists() ) {
+	public Optional<Path> findProjectRoot( final Path base ) {
+		if ( base == null || !base.toAbsolutePath().toFile().exists() ) {
```
diba1013, line 24: Does `Files.exist(Path)` not work?

src/main/java/de/retest/recheck/configuration/ProjectConfigurationUtil.java
```diff
@@ -7,15 +7,23 @@
...
-	public static Optional<Path> findProjectConfiguration( final Path path ) {
-		return ProjectRootFinderUtil.getProjectRoot( path ).map( ProjectConfigurationUtil::resolveProperties );
+	public static Optional<Path> findProjectConfigurationFolder( final Path basePath ) {
+		return ProjectRootFinderUtil.getProjectRoot( basePath )
+				.map( path -> path.resolve( ProjectConfiguration.RETEST_PROJECT_CONFIG_FOLDER ) );
 	}
 
-	private static Path resolveProperties( final Path path ) {
-		return path.resolve( ProjectConfiguration.RETEST_PROJECT_PROPERTIES );
+	public static Optional<Path> findProjectConfigurationProperties() {
```
diba1013: I would rename this to `findProjectProperties` to not confuse with the configuration folder.

diba1013 (changes requested):

beatngu13 (approved):

diba1013 (approved):

PR #123, opened 2019-02-27
Improve options

beatngu13:


beatngu13:
While this:

```java
RecheckOptions opts = new RecheckOptions();
opts.setFoo( "Bar" );
Recheck re = new RecheckImpl( opts );
```

Is closer to options in Selenium, I would personally prefer immutable options with a (Lombok-style) builder:

```java
RecheckOptions opts = RecheckOptions.builder()
        .foo( "Bar" )
        .build();
Recheck re = new RecheckImpl( opts );
```

modulo11 (approved):

PR #124, opened 2019-02-28
Add a gradle compatible file namer

narkaTee:
A file namer which is compatible with the default gradle folder layout.

Currently it uses the hardcoded test reports path `build/test-results/test` for the result files.
The workspace path is identical to the maven file namer.

This will work with the default gradle sourceSet layout for java projects.
Gradle allows users to reconfigure the sourceSets (adding new, etc.), so this might break depending how "strange" die build config is.
I will test some scenarios and report back with the results.

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -8,4 +8,8 @@ set -o pipefail
 export MVN="mvn --settings ${TRAVIS_BUILD_DIR}/.travis.settings.xml"
 
 ###### Maven ######
-${MVN} clean org.jacoco:jacoco-maven-plugin:prepare-agent verify sonar:sonar
+if [ ${TRAVIS_SECURE_ENV_VARS} = "true" ]; then
+    ${MVN} clean org.jacoco:jacoco-maven-plugin:prepare-agent verify sonar:sonar
+else
+    ${MVN} clean org.jacoco:jacoco-maven-plugin:prepare-agent verify
+fi
--- a/src/main/java/de/retest/recheck/FileNamerStrategy.java
+++ b/src/main/java/de/retest/recheck/FileNamerStrategy.java
@@ -2,10 +2,27 @@ package de.retest.recheck;
 
 import de.retest.recheck.persistence.FileNamer;
 
+/**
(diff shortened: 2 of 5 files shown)


beatngu13:
First of all, thanks for the PR! I've made some adaptions here:

https://github.com/retest/recheck/compare/feature/gradle-file-namer

I've also added a test where I included the expected results. I'm not that familiar with Gradle, but I think

```
src/test/resources/resources/retest/foo/bar.recheck
build/test-results/test/foo/bar.result
```

should be

```
src/test/resources/resources/retest/recheck/foo/bar.recheck
build/test-results/test/retest/recheck/foo/bar.result
```

?

Also, feel free to include yourself in `<contributors>` within the POM.

beatngu13 (changes requested):
See feature branch in recheck.

narkaTee:
> First of all, thanks for the PR! I've made some adaptions here:
> 
> https://github.com/retest/recheck/compare/feature/gradle-file-namer

I have added support for different gradle source sets, by allowing a source set name to passed to the namer.



> I've also added a test where I included the expected results. I'm not that familiar with Gradle, but I think
> 
> ```
> src/test/resources/resources/retest/foo/bar.recheck
> build/test-results/test/foo/bar.result
> ```
> should be
> 
> ```
> src/test/resources/resources/retest/recheck/foo/bar.recheck
> build/test-results/test/retest/recheck/foo/bar.result
> ```
> ?

Gradle only creates the folders up to `build/test-results/test/`, every folder after that has to be created by retest.

I was actually confused about who creates the folder structure `target/test-classes/retest/recheck/`, I assumed maven does that because I could not find a code path in `RecheckReplayResultUtil` that creates folders.
That is why I choose `build/test-results/test/`.

But I would also prefer `build/test-results/test/retest/recheck/`.
Maybe `RecheckReplayResultUtil.persist` could use `FileUtil.ensureFolder` to create the folders?

beatngu13 (approved):

beatngu13:
Thx @narkaTee! 🙏

PR #134, opened 2019-03-06
Do not mark if element should be ignored

roesslerj:


src/main/java/de/retest/recheck/ui/diff/ElementDifference.java
```diff
@@ -48,15 +48,18 @@ public ElementDifference( final Element element, final AttributesDifference attr
 		this.childDifferences.addAll( childDifferences );
 	}
 
-	public Screenshot mark( final Screenshot screenshot ) {
+	public Screenshot mark( final Screenshot screenshot, final ShouldIgnore ignore ) {
 		if ( screenshot == null ) {
 			return null;
 		}
 		final List<Rectangle> marks = new ArrayList<>();
```
beatngu13, line 55: Just passing by here… 😊 You could also use a stream, something like:

```java
final List<Rectangle> marks = childDifferences == null ? Collections.emptyList() : childDifferences.stream() //
		.flatMap( diff -> diff.getNonEmptyDifferences().stream() ) //
		.filter( elementDiff -> !ignore.shouldIgnoreElement( element ) ) //
		.filter( elementDiff -> !elementDiff.getAttributeDifferences( ignore ).isEmpty() ) //
		.map( elementDiff -> AttributeUtil.getAbsoluteOutline( elementDiff.getIdentifyingAttributes() ) ) //
		.collect( Collectors.toList() );
```

roesslerj: But this is optional, right. I mean stream-API is not the new `must use`.

beatngu13 (approved):

src/main/java/de/retest/recheck/ui/diff/ElementDifference.java
beatngu13: That's why I said "you *could*". It was just a comment, take it easy. 😉

PR #142, opened 2019-03-12
Improve default files and point to documentation

roesslerj:


src/main/resources/default-recheck.ignore
```diff
@@ -1 +1,13 @@
...
+# matcher: retestid=div-b4f23, attribute: font.*
+
+# To ignore whole subtrees, use e.g. xpath:
+# matcher: xpath=HTML[1]/BODY[1]/DIV[1]/DIV[1]/DIV[1]
+
+# To ignore attributes globally, use 
+# attribute=class
+
+# More details and examples can be found here:
+# https://github.com/retest/docs/blob/master/docs/recheck/how-ignore-works-in-recheck.md
```
beatngu13: I would suggest to always link against the deployed documentation: https://retest.github.io/docs/recheck/how-ignore-works-in-recheck/.

src/main/resources/default-recheck.ignore.js
```diff
@@ -1 +1,9 @@
-// Ignore for recheck. Please do not delete, even if they appear to be empty
+// Here you can implement ignore rules for recheck in javascript.
+// Please do not delete this file, even if it is empty.
+
+// You can implement either of these two functions:
+// shouldIgnoreElement(element)
+// shouldIgnoreAttributeDifference(element, diff)
+
+// You can find more details and example rules at: 
+// https://github.com/retest/docs/blob/master/docs/recheck/how-ignore-works-in-recheck.md
```
beatngu13: Here too: https://retest.github.io/docs/recheck/how-ignore-works-in-recheck/.

src/main/resources/default-recheck.ignore
roesslerj: Yes please, and it's also way shorter...

src/main/resources/default-recheck.ignore
```diff
@@ -1 +1,13 @@
...
+# matcher: retestid=div-b4f23, attribute: font.*
+
+# To ignore whole subtrees, use e.g. xpath:
+# matcher: xpath=HTML[1]/BODY[1]/DIV[1]/DIV[1]/DIV[1]
+
+# To ignore attributes globally, use 
+# attribute=class
+
+# More details and examples can be found here:
+# https://retest.github.io/docs/recheck/how-ignore-works-in-recheck/
```
modulo11, line 13: Maybe we should just link to the main page.

roesslerj: In case we have a page for a specific topic ... why not link to it?

beatngu13 (approved):

PR #148, opened 2019-03-13
Fix enum name

beatngu13:


beatngu13:
Is just wonder: Is this how we are supposed to deal with a renamed class that was previously supported by XML migrators?

roesslerj:
I do not think that makes sense. Let's talk about this tomorrow...

PR #154, opened 2019-03-21
Fix ignore loading

beatngu13:
I've reverted fc48b8a63ccb0c2c7d5234a0a1600431534f19ad because `Files.lines(…)` in `GlobalIgnoreApplier.load()` doesn't read in the terminating empty line break at the end, which would again cause a diff in the VCS. However, I think this is already an improvement.


src/test/java/de/retest/recheck/review/workers/LoadShouldIgnoreWorkerTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+		System.setProperty( RETEST_PROJECT_ROOT, configFolder.toString() );
+	}
+
+	@Test
+	@SystemProperty( key = RETEST_PROJECT_ROOT )
+	void loaded_ignore_file_should_match_default_ignore_file() throws Exception {
+		final LoadShouldIgnoreWorker cut = new LoadShouldIgnoreWorker( NopCounter.getInstance() );
+		final GlobalIgnoreApplier globalIgnoreApplier = cut.load();
+		final List<String> loadedIgnoreFileLines = globalIgnoreApplier.persist().getIgnores().stream() //
```
diba1013: Couldn't you just load and save via `SaveShouldIgnoreWorker` and check if the files are equal?
`assertThat( file ).hasSameContentAs( other )`

src/main/java/de/retest/recheck/review/ignore/ShouldIgnoreEmptyLineLoader.java
```diff
@@ -0,0 +1,54 @@
...
+
+import de.retest.recheck.ignore.ShouldIgnore;
+import de.retest.recheck.review.ignore.ShouldIgnoreEmptyLineLoader.ShouldIgnoreEmptyLine;
+import de.retest.recheck.review.ignore.io.Loader;
+import de.retest.recheck.ui.descriptors.Element;
+import de.retest.recheck.ui.diff.AttributeDifference;
+
+public class ShouldIgnoreEmptyLineLoader implements Loader<ShouldIgnoreEmptyLine> {
+
+	public static class ShouldIgnoreEmptyLine implements ShouldIgnore {
```
diba1013: This is the same as `ShouldIgnoreComment` should we merge these?

src/main/java/de/retest/recheck/review/ignore/io/Loaders.java
```diff
@@ -72,9 +74,7 @@ private Loaders() {}
 	}
 
 	public static Stream<?> load( final Stream<String> lines ) {
-		return lines.map( String::trim ) //
-				.filter( StringUtils::isNotEmpty ) //
-				.map( Loaders::load );
+		return lines.map( Loaders::load );
```
diba1013, line 74: We need to take care that users know that they should not indent their rules, else there will be an exception. Or we could add a loader that, in the end, tries to trim and delegates to another loader.

src/main/resources/default-recheck.ignore
```diff
@@ -10,4 +10,4 @@
 # attribute=class
 
 # More details and examples can be found here:
-# https://retest.github.io/docs/recheck/how-ignore-works-in-recheck/
+# https://retest.github.io/docs/recheck/how-ignore-works-in-recheck/
```
diba1013: Any comment on why revert. Seems like the stackoverflow answer is reasonable.

diba1013 (changes requested):

beatngu13:
I think I have now also resolved the file ending issue. `JSShouldIgnoreLoader` has:

```java
@Override
public String save( final JSShouldIgnoreImpl ignore ) {
	return "";
}
```

Which creates unnecessary file content. I now filter it on save.

src/main/java/de/retest/recheck/review/workers/SaveShouldIgnoreWorker.java
```diff
@@ -21,13 +24,16 @@ public SaveShouldIgnoreWorker( final GlobalIgnoreApplier applier ) {
...
 
 	public void save() throws IOException {
-		final Optional<Path> path = RecheckIgnoreUtil.getIgnoreFile();
+		final Optional<Path> ignoreFile = RecheckIgnoreUtil.getIgnoreFile();
 		final PersistableGlobalIgnoreApplier persist = applier.persist();
 
-		final Stream<String> save = Loaders.save( persist.getIgnores().stream() );
+		// Filter JSShouldIgnore because that would create unnecessary file content.
+		final Stream<ShouldIgnore> ignores = persist.getIgnores().stream() //
+				.filter( ((Predicate<ShouldIgnore>) JSShouldIgnoreImpl.class::isInstance).negate() );
```
diba1013: Which is more readable?
```suggestion
				.filter( (ignore -> !(ignore instanceof JSShouldIgnoreImpl) );
```

src/main/java/de/retest/recheck/review/ignore/ShouldIgnorePreserveLineLoader.java
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		final boolean onlyWhiteSpace = StringUtils.isWhitespace( line );
+		if ( onlyWhiteSpace ) {
+			return true;
+		}
+
+		final boolean leadingWhitespace = Character.isWhitespace( line.charAt( 0 ) );
+		if ( leadingWhitespace ) {
+			log.warn( "Please remove leading whitespace from the following line:" );
```
diba1013: Just minor. If we have multiple parallel log output, this line would be lost. I would merge this into one line.
```suggestion
			log.warn( "Please remove leading whitespace from the following line:\n{}", line );
```

src/main/java/de/retest/recheck/review/ignore/ShouldIgnorePreserveLineLoader.java
```diff
@@ -0,0 +1,75 @@
...
+			log.warn( "{}", line );
+			return true;
+		}
+
+		return false;
```
diba1013, line 61: Since this loader is the last one anyways, should we always preserve lines, but add a warning (then we need to ensure that this is last, e.g. add a comment)? Otherwise there is an exception thrown.

diba1013 (changes requested):

src/main/java/de/retest/recheck/review/ignore/ShouldIgnorePreserveLineLoader.java
beatngu13: Although it is true that an exception is thrown, the last loader is actually `JSShouldIgnoreImpl`. We should probably remove `JSShouldIgnoreImpl` from here and also change the default implementation of `canLoad( String )`.

However, I think this belongs in a different PR, respectively, multiple PRs, not this one.

diba1013 (approved):

PR #155, opened 2019-03-21
Max pixel diff

beatngu13:
Draft for a general approach to ignore pixel diffs. Will wait for the next innovations & ideas meeting, but what do you think? Usage is quite simple:

```java
RecheckOptions opts = RecheckOptions.builder().maxPixelDiff(5.0).build();
Recheck re = new RecheckImpl(opts);
```


diba1013:
I would make this configurable i.e. via ignore. I've got some ideas coming up in my workshop. We might need therefore https://github.com/retest/recheck-web/issues/182.

beatngu13:
> I would make this configurable i.e. via ignore.

Since ignores are additive, you can use both. We already use a similar approach:

https://github.com/retest/recheck-web/blob/master/.retest/recheck.ignore.js

>I've got some ideas coming up in my workshop. We might need therefore [retest/recheck-web#182](https://github.com/retest/recheck-web/issues/182).

Sounds good! 👍

beatngu13:
Will open another draft with an ignore-based solution.

PR #162, opened 2019-04-02
Max pixel diff ignore

beatngu13:
Draft we can discuss tomorrow in our innovation meeting. Basically, this allows to ignore max pixel diffs by adding e.g. `maxPixelDiff=5` (or `5.0`) to `recheck.ignore`.

### DIFF
--- a/src/main/java/de/retest/recheck/review/ignore/MaxPixelDiffShouldIgnore.java
+++ b/src/main/java/de/retest/recheck/review/ignore/MaxPixelDiffShouldIgnore.java
@@ -1,0 +1,99 @@
+package de.retest.recheck.review.ignore;
+
+import java.awt.Rectangle;
+import java.io.Serializable;
+import java.util.regex.MatchResult;
+import java.util.regex.Pattern;
+
+import de.retest.recheck.ignore.ShouldIgnore;
+import de.retest.recheck.review.ignore.io.RegexLoader;
+import de.retest.recheck.ui.descriptors.Element;
+import de.retest.recheck.ui.diff.AttributeDifference;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class MaxPixelDiffShouldIgnore implements ShouldIgnore {
+
+	private static final String PIXEL = "px";
+
+	private final double maxPixelDiff;
+
+	public MaxPixelDiffShouldIgnore( final double maxPixelDiff ) {
+		this.maxPixelDiff = maxPixelDiff;
+	}
+
+	@Override
+	public boolean shouldIgnoreElement( final Element element ) {
+		return false;
+	}
+
+	@Override
+	public boolean shouldIgnoreAttributeDifference( final Element element,
+			final AttributeDifference attributeDifference ) {
+		final Serializable expected = attributeDifference.getExpected();
+		final Serializable actual = attributeDifference.getActual();
+
+		if ( expected instanceof Rectangle ) {
+			return checkRectangle( (Rectangle) expected, (Rectangle) actual );
+		}
+
+		if ( expected instanceof String ) {
+			return checkString( (String) expected, (String) actual );
+		}
+
+		return false;
+	}
+
+	private boolean checkRectangle( final Rectangle expected, final Rectangle actual ) {
+		final boolean ignoreX = Math.abs( expected.x - actual.x ) <= maxPixelDiff;
+		final boolean ignoreY = Math.abs( expected.y - actual.y ) <= maxPixelDiff;
+		final boolean ignoreHeight = Math.abs( expected.height - actual.height ) <= maxPixelDiff;
+		final boolean ignoreWidth = Math.abs( expected.width - actual.width ) <= maxPixelDiff;
+		return ignoreX && ignoreY && ignoreHeight && ignoreWidth;
+	}
+
+	private boolean checkString( final String expected, final String actual ) {
+		if ( expected == null || actual == null ) {
+			return false;
+		}
+
+		if ( !expected.endsWith( PIXEL ) || !actual.endsWith( PIXEL ) ) {
+			return false;
+		}
+
+		try {
+			final double expectedDouble = Double.parseDouble( clean( expected ) );
+			final double actualDouble = Double.parseDouble( clean( actual ) );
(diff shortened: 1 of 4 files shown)


src/main/java/de/retest/recheck/review/ignore/MaxPixelDiffShouldIgnore.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.regex.Pattern;
+
+import de.retest.recheck.ignore.ShouldIgnore;
+import de.retest.recheck.review.ignore.io.RegexLoader;
+import de.retest.recheck.ui.descriptors.Element;
+import de.retest.recheck.ui.diff.AttributeDifference;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class MaxPixelDiffShouldIgnore implements ShouldIgnore {
```
diba1013, line 15: This is really a web-feature, isn't it? I would move this to recheck-web. However we then need to have a extensible way of adding Loaders.

src/main/java/de/retest/recheck/review/ignore/MaxPixelDiffShouldIgnore.java
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	@Override
+	public String toString() {
+		return String.format( MaxPixelDiffShouldIgnoreLoader.FORMAT, maxPixelDiff );
+	}
+
+	public static class MaxPixelDiffShouldIgnoreLoader extends RegexLoader<MaxPixelDiffShouldIgnore> {
+
+		private static final String KEY = "maxPixelDiff=";
```
diba1013, line 85: We need to take care of the syntax. Since this can potentially ignore too much: `outline`, `margin-*`, `padding-*` and thus  potentially ignoring unexpected differences.

diba1013: What was your use-case here?

diba1013: When I thought of improving the syntax, I thought of something like:
`attribute: $key, (value|expect|actual)=$val` where
* value targets both expected or actual values
* expect targets only expected values
* actual targets only actual values
* $val is *equal* comparison

If you need something more configurable, you could do:
* `attribute: $key, (value|expect|actual): $expression` where `$expression` can be
    * `eq=$val`
    * `gt=$val`
    * ...
    * `between=$range`
    * `within=(+-)?$range`
* $range being a classical interval: `()`, `(]`, `[)` and `[]` or a direct value

We need to keep in mind that:
* `=` is a direct value
* `:` is a delegate to another matcher

Examples:
`attribute: margin-bottom, value: between=[1,2]`
`attribute: outline, expect: within=+-5`

Let's talk in the office more about that.

src/main/java/de/retest/recheck/review/ignore/MaxPixelDiffShouldIgnore.java
```diff
@@ -0,0 +1,99 @@
...
+		private static final String FORMAT = KEY + "%s";
+		private static final Pattern REGEX = Pattern.compile( KEY + "(\\d+(\\.\\d+)?)" );
```
diba1013, line 87: Some people are used to do `.5` meaning `0.5`

diba1013 (changes requested):

diba1013:
Is documentation on its way?

src/main/java/de/retest/recheck/review/ignore/MaxPixelDiffShouldIgnore.java
beatngu13: I'm not sure. While we currently only face issues with pixel diffs in recheck-web, I can't tell if such a feature wouldn't also be interesting for Swing, Android, or something else.

As you already mentioned, we ain't got no extensible way of adding loaders. Therefore, I would keep it in recheck.

src/main/java/de/retest/recheck/review/ignore/MaxPixelDiffShouldIgnore.java
beatngu13: It is up to us if we want to allow this, I just thought 5 and 5.0 is sufficient.

beatngu13:
> Is documentation on its way?

No, because I have marked it as a draft. I first want to discuss the solution with the team. As soon as we have a final design, documention will have to follow.

src/main/java/de/retest/recheck/review/ignore/MaxPixelDiffShouldIgnore.java
beatngu13: >What was your use-case here?

IMHO ignoring small rendering differences among different browsers and/or target devices.

I will propose the solution later on and then we can improve it.

Also:

>We need to keep in mind that:
>
>* = is a direct value
>* : is a delegate to another matcher

This wasn't clear to me as I haven't found such information in the docs:

https://retest.github.io/docs/recheck/how-ignore-works-in-recheck/

Maybe we should add this?

beatngu13: Again regarding the use case, by Kent C. Odds (JavaScript engineer at PayPal and creator of TestingJavaScript.com) via applitools.com:

>I love how Applitools Eyes has the smarts to ignore minor visual differences in your components, especially when testing on different browser types where it’s easy for layout to be off by a pixel.

I think a concise and easy way to do this is very helpful. Currently, this can only be done with a bunch of JS: https://github.com/retest/recheck-web/blob/master/.retest/recheck.ignore.js

beatngu13:
TODOs:

* Exclude font(s)
* What else to exclude?
* Related to RET-1636?

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

PR #163, opened 2019-04-03
Feature/fix sutstate null

alig01:
As soon as createActionChangeSet does not get passed the stateFilePath, an exception can be thrown. Therefore, the StateFilePath should be handed over to avoid the error.

### DIFF
--- a/src/test/java/de/retest/suite/flow/CreateChangesetForAllDiffrencesFlowTest.java
+++ b/src/test/java/de/retest/suite/flow/CreateChangesetForAllDiffrencesFlowTest.java
@@ -1,0 +1,104 @@
+package de.retest.suite.flow;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.io.File;
+import java.nio.file.Files;
+import java.util.Arrays;
+import java.util.HashSet;
+
+import org.apache.commons.io.FileUtils;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import de.retest.recheck.configuration.ProjectConfiguration;
+import de.retest.recheck.persistence.NoGoldenMasterFoundException;
+import de.retest.recheck.persistence.Persistence;
+import de.retest.recheck.persistence.PersistenceFactory;
+import de.retest.recheck.persistence.xml.util.StdXmlClassesProvider;
+import de.retest.recheck.suite.flow.ApplyChangesToStatesFlow;
+import de.retest.recheck.ui.descriptors.Element;
+import de.retest.recheck.ui.descriptors.SutState;
+import de.retest.recheck.ui.diff.AttributeDifference;
+import de.retest.recheck.ui.review.ReviewResult;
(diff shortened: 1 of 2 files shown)


beatngu13:
@IgorAlbert We can either write a test or temporarily disable the quality gate, respectively, the status check here on GitHub.

src/test/java/de/retest/suite/flow/CreateChangesetForAllDiffrencesFlowTest.java
```diff
@@ -0,0 +1,100 @@
...
+		Files.createDirectories( temp.getRoot().toPath().resolve( "src/test/java" ) );
+	}
+
+	@Test
+	public void apply_should_throw_exception_whitout_the_stateFilePath() throws Exception {
+		final PersistenceFactory persistenceFactory =
+				new PersistenceFactory( new HashSet<>( Arrays.asList( StdXmlClassesProvider.getXmlDataClasses() ) ) );
+
+		final Persistence<SutState> persistence = persistenceFactory.getPersistence();
+		final File originalFile = new File( FILE );
```
beatngu13, line 55: Why not move the entire file copying and so forth into `setup()`?

src/test/java/de/retest/suite/flow/CreateChangesetForAllDiffrencesFlowTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+		//if createActionChangeSet is called whitout the stateFilePath, a NoStateFileFoundException is thrown
+		changesWhitoutStateFilePath.createTestChangeSet().createActionChangeSet( ACTION, null ).getAttributesChanges()
+				.add( toChange.getIdentifyingAttributes(), new AttributeDifference( "enabled", "false", "true" ) );
+
+		changesWhitoutAnything.createTestChangeSet().createActionChangeSet().getAttributesChanges()
+				.add( toChange.getIdentifyingAttributes(), new AttributeDifference( "enabled", "false", "true" ) );
+
+		assertThatCode( () -> ApplyChangesToStatesFlow.apply( persistence, changesWhitPathAndDescription ) )
+				.doesNotThrowAnyException();
```
beatngu13, line 90: We usually don't use this assertion because if a particular statement throws an exception, the test fails anyway. However, is this critical? I don't know the story behind it.

beatngu13 (changes requested):

src/test/java/de/retest/suite/flow/CreateChangesetForAllDiffrencesFlowTest.java
alig01: Jeremy had a problem that this NoStateFileFoundException was thrown, when he had tested something. He solved it with the other constructor, where the 2 parameters are passed, but he could not verify to 100% that this will fix the problem.
So I wanted to prove that the exception is not thrown when stateFilePath is passed.

src/test/java/de/retest/suite/flow/CreateChangesetForAllDiffrencesFlowTest.java
alig01: Done

beatngu13 (dismissed):

beatngu13 (approved):

PR #164, opened 2019-04-04
Feature/match attributes

roesslerj:


src/test/resources/de/retest/recheck/review/workers/recheck.ignore
```diff
@@ -2,6 +2,7 @@ matcher: id=banner
 matcher: id=title, attribute: font
 matcher: retestid=banner
 matcher: retestid=banner, attribute: outline
+matcher: retestid=banner, attribute: .*color
```
diba1013: What if i want to ignore a custom html-attribute: `data.theme` but not `data-theme`. We then would have to remember to escape this `data\.theme`, which is complex. 

I would rather use https://github.com/retest/recheck/pull/162#discussion_r271596388 with `$expression: regex=.*color`. Then it would be clear, that this needs escaping.

Regardless, we should be consistent with the `equals` or `matches` so that the user does not need to remember 1000 different use cases, but instead 5.

As proposed yesterday, do we need a workshop for this?

diba1013 (changes requested):

src/test/resources/de/retest/recheck/review/workers/recheck.ignore
roesslerj: I don't think a workshop is necessary. I'll just add Daniel as reviewer...

beatngu13:
I like the adapted version using `attribute-regex`, very clear, no escaping needed in `attribute`. So if @diba1013 is OK with this, I would approve as soon as we have documentation for this.

beatngu13:
As discussed, we should at least have a branch in docs that updates the current documention:

https://retest.github.io/docs/recheck/how-ignore-works-in-recheck/

According to this, one can already use regex.

beatngu13 (approved):

diba1013 (approved):

PR #166, opened 2019-04-05
Feature/#159 fix terminology

### LINKED ISSUE #159: Unify terminology
Besides "golden master"—the term we agreed on—we use different words to describe the same thing. For instance:

* `RecheckStateFileProviderImpl`: "state file"
* `NoRecheckFileActionReplayResult`: "recheck file"
* …

I think it would be helpful for both the user and us to unify the terminology and consistently use the term "golden master".

enesoezel:
resolves#159

### DIFF
--- a/src/main/java/de/retest/recheck/NoGoldenMasterActionReplayResult.java
+++ b/src/main/java/de/retest/recheck/NoGoldenMasterActionReplayResult.java
@@ -20,9 +20,9 @@ import de.retest.recheck.ui.diff.StateDifference;
 
 public class NoGoldenMasterActionReplayResult extends ActionReplayResult {
 
-	private static final String MSG_SHORT = "No recheck file found.";
+	private static final String MSG_SHORT = "No Golden Master found.";
 	public static final String MSG_LONG =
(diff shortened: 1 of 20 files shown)


src/main/java/de/retest/recheck/NoGoldenMasterActionReplayResult.java
```diff
@@ -20,9 +20,9 @@
 
 public class NoGoldenMasterActionReplayResult extends ActionReplayResult {
 
-	private static final String MSG_SHORT = "No recheck file found.";
+	private static final String MSG_SHORT = "No 'Golden Master' found.";
```
beatngu13: Needs no (single) quotes here and elsewhere.

beatngu13 (changes requested):
Looks good. I just wonder if this breaks compatability beyond our own projects, i.e., external API? @modulo11 @roesslerj

diba1013:
Just a reminder: Documentation

beatngu13:
> Just a reminder: Documentation

We already use Golden Master there, but we probably should explain what is the difference between a Golden Master and a (SUT) state.

beatngu13 (approved):

PR #170, opened 2019-04-11
Increase similarity needed for similarity as been shown customer project

roesslerj:


src/main/java/de/retest/recheck/image/ImageDifference.java
```diff
@@ -4,7 +4,7 @@
 
 public class ImageDifference {
 	// to make sure to account for rounding errors
-	private static final double EQUALITY_THRESHOLD = 0.99;
+	private static final double EQUALITY_THRESHOLD = 0.9999999999;
```
alig01, line 7: is the number chosen arbitrarily?

roesslerj: I talked with Martin about that... the number _is_ chosen arbitrarily, but until we have more real-life examples, we probably want to keep it that way... :-/

alig01 (approved):

PR #171, opened 2019-04-11
#RET 1635 Create visibility filter file

ivonx:


src/main/resources/filter/web/visibility.filter
```diff
@@ -0,0 +1,136 @@
...
+attribute=object-fit
+attribute=opacity
+attribute=order
+attribute=outline-color
+attribute=outline-style
+attribute=page-break-after
+attribute=page-break-before
+attribute=page-break-inside
+attribute=perspective
+  #  - perspective-origin
```
beatngu13: Shouldn't this be:

```
#attribute=perspective-origin
```

Copy & paste from `attributes.yaml`? 😉

beatngu13 (dismissed):
Same as in https://github.com/retest/recheck/pull/172.

src/main/resources/filter/web/visibility.filter
```diff
@@ -0,0 +1,188 @@
+# Visibility filter file for recheck.
+
+# Ignore visibility CSS attributes:
+
+attribute=align-content
+attribute=align-items
+attribute=align-self
+attribute=all
```
roesslerj: I think `all` is a retest internal attribute. I don't think that is an existing CSS attribute?

src/main/resources/filter/web/visibility.filter
```diff
@@ -0,0 +1,188 @@
...
+attribute=cols
+attribute=colspan
+attribute=content
+attribute=high
+attribute=href
+attribute=hreflang
+attribute=kind
+attribute=label
+attribute=type
+attribute=value
```
roesslerj: I would add `src` and newline at end of file.

roesslerj (changes requested):

roesslerj (approved):

roesslerj (approved):

PR #177, opened 2019-04-18
Feature/improve error msgs

roesslerj:


src/main/java/de/retest/recheck/execution/RecheckAdapters.java
```diff
@@ -9,16 +9,23 @@
...
 	public static RecheckAdapter findAdapterFor( final Object toVerify ) {
 		return StreamSupport.stream( adapters.spliterator(), false ) //
 				.filter( adapter -> adapter.canCheck( toVerify ) ) //
 				.findAny() //
-				.orElseThrow( () -> new UnsupportedOperationException( String.format(
-						"No recheck adapter registered that can handle an object of %s.", toVerify.getClass() ) ) );
+				.orElseThrow( () -> createHelpfulExceptionForMissingAdapter( toVerify.getClass().getName() ) );
+	}
+
+	protected static UnsupportedOperationException createHelpfulExceptionForMissingAdapter( final String className ) {
```
beatngu13: Good workaround for mocking `getClass()`! 👍 

You could omit the access modifier as package private is sufficient.

src/main/java/de/retest/recheck/execution/RecheckAdapters.java
```diff
@@ -9,16 +9,23 @@
...
+		final String msg =
+				String.format( "No recheck adapter registered that can handle an object of class %s.", className );
+		if ( className.startsWith( "org.openqa.selenium" ) ) {
+			return new UnsupportedOperationException(
+					msg + "\n You need to add recheck-web (https://github.com/retest/recheck-web) to the classpath." );
```
beatngu13, line 26: Maybe "Please use recheck-web (…) instead of recheck directly."? But we can also keep it as it is.

src/main/java/de/retest/recheck/RecheckImpl.java
```diff
@@ -45,6 +58,11 @@
 	private final TestReplayResultPrinter printer;
 	private final GlobalIgnoreApplier ignoreApplier;
 
+	/**
+	 * Constructor that works purely with defaults. Assumes being called from within a JUnit test. If you want to use
```
beatngu13: Hmmm, why JUnit? AFAIK our [`TestCaseFinder`](https://github.com/retest/recheck/blob/master/src/main/java/de/retest/recheck/TestCaseFinder.java) also supports TestNG.

src/main/java/de/retest/recheck/RecheckImpl.java
```diff
@@ -27,6 +27,19 @@
 import de.retest.recheck.ui.descriptors.SutState;
 import de.retest.recheck.ui.diff.LeafDifference;
 
+/**
+ * Base implementation to use with e.g. recheck-web (https://github.com/retest/recheck-web) or any other recheck
+ * implementation. Defines the principle test API and implements functionality on the basis of interfaces. <br />
+ * Principle usage pattern (in the context of JUnit):
```
beatngu13: JUnit and TestNG?

beatngu13 (changes requested):

beatngu13:
There is also a build failure due to Javadoc: https://travis-ci.com/retest/recheck/builds/108825345

src/main/java/de/retest/recheck/execution/RecheckAdapters.java
roesslerj: Since it is possible, that this problem arises also when combining multiple rechecks (e.g. XML, web), I would keep it as is.

beatngu13:
SonarCloud complains about the generic `RuntimeException`s. I think a dedicated exception would be overengineered, but does a `NullPointerException` makes sense or shall we mark the issues as "won't fix"?

beatngu13 (approved):

PR #187, opened 2019-04-24
Feature/ret 1638 support customer filter

alig01:
Since the return type of getProjectFilterFiles() has been changed to List<Path>, the other method getDefaultFilterFiles() must also be changed with a different PR

### DIFF
--- a/src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
+++ b/src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
@@ -1,5 +1,7 @@
 package de.retest.recheck.ignore;
 
+import static de.retest.recheck.configuration.ProjectConfiguration.FILTER_FOLDER;
+
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -11,6 +13,8 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import de.retest.recheck.configuration.ProjectConfiguration;
+
 public class SearchFilterFiles {
 
 	public static final String FILES_ENDING = ".filter";
@@ -20,6 +24,10 @@ public class SearchFilterFiles {
 			Arrays.asList( WEB_FILTER_DIR + "positioning.filter", WEB_FILTER_DIR + "visibility.filter" );
 
 	private SearchFilterFiles() {}
+
+	private static Path resolveFilterPath() {
+		return ProjectConfiguration.getInstance().findProjectConfigFolder().resolve( FILTER_FOLDER );
+	}
 
 	public static List<File> getDefaultFilterFiles() {
 		return defaultWebFilter.stream() //
--- a/src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
+++ b/src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
@@ -1,13 +1,19 @@
 package de.retest.recheck.ignore;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class SearchFilterFiles {
 
+	public static final String FILES_ENDING = ".filter";
 	private static final String BASIC_FILTER_DIR = "/filter/";
 	private static final String WEB_FILTER_DIR = BASIC_FILTER_DIR + "web/";
 	private static final List<String> defaultWebFilter =
@@ -23,4 +29,15 @@ public class SearchFilterFiles {
 				.collect( Collectors.toList() ); //
 	}
 
+	public static List<File> getProjectFilterFiles() throws IOException {
+		if ( !resolveFilterPath().toFile().exists() ) {
+			return Collections.emptyList();
+		}
+		try ( Stream<Path> paths = Files.walk( resolveFilterPath() ) ) {
+			return paths.filter( Files::isRegularFile ) //
+					.filter( file -> file.toString().endsWith( FILES_ENDING ) ) //
+					.map( Path::toFile ) //
+					.collect( Collectors.toList() ); //
+		}
+	}
 }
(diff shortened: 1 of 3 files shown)


src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -15,6 +26,14 @@
 
 	private SearchFilterFiles() {}
 
+	private static Path resolveFilterPath() {
```
martin-v: This method duplicates code from `ProjectConfiguration.ensureProjectConfigurationInitialized()`, please refactore.

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -23,4 +42,15 @@ private SearchFilterFiles() {}
...
 	}
 
+	public static List<File> getProjectFilterFiles() {
+		try ( Stream<Path> paths = Files.walk( resolveFilterPath() ) ) {
+			return paths.filter( Files::isRegularFile ) //
+					.filter( file -> file.toString().endsWith( ".filter" ) ) //
+					.map( Path::toFile ) //
+					.collect( Collectors.toList() ); //
+		} catch ( final Exception e ) {
+			log.info( "You can create your own filter files in the \\.retest\\filter folder" );
```
martin-v: Use `File.SEPERATOR` instead of OS dependent `\`

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -23,4 +42,15 @@ private SearchFilterFiles() {}
...
+		}
+		return new ArrayList<>();
```
martin-v: Move `return` in the `catch` block because it is only for this case, then the code is better readable.

martin-v (dismissed):

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -23,4 +37,19 @@ private SearchFilterFiles() {}
...
+	public static List<File> getProjectFilterFiles() {
+		if ( !resolveFilterPath().toFile().exists() ) {
+			ProjectConfiguration.getInstance().ensureProjectConfigurationInitialized();
+		}
+		try ( Stream<Path> paths = Files.walk( resolveFilterPath() ) ) {
+			return paths.filter( Files::isRegularFile ) //
+					.filter( file -> file.toString().endsWith( ".filter" ) ) //
+					.map( Path::toFile ) //
+					.collect( Collectors.toList() ); //
+		} catch ( final Exception e ) {
```
diba1013: Please, for once, pass the exception upwards...

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -23,4 +37,19 @@ private SearchFilterFiles() {}
...
+			log.info( "You can create your own filter files in the " + File.separator + ".retest" + File.separator
+					+ "filter folder" );
```
diba1013: You are checking that the directory exists and creating it if necessary. You should not wait for an exception to be thrown. This should be part of the documentation, not the code.

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -23,4 +37,19 @@ private SearchFilterFiles() {}
...
+			return new ArrayList<>();
```
diba1013: Always, `Collections.emptyList()` for empty lists.

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -23,4 +37,19 @@ private SearchFilterFiles() {}
 				.collect( Collectors.toList() ); //
 	}
 
+	public static List<File> getProjectFilterFiles() {
+		if ( !resolveFilterPath().toFile().exists() ) {
+			ProjectConfiguration.getInstance().ensureProjectConfigurationInitialized();
+		}
+		try ( Stream<Path> paths = Files.walk( resolveFilterPath() ) ) {
+			return paths.filter( Files::isRegularFile ) //
+					.filter( file -> file.toString().endsWith( ".filter" ) ) //
```
diba1013: Do we have a constant anywhere for `.filter`?

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -23,4 +37,19 @@ private SearchFilterFiles() {}
 				.collect( Collectors.toList() ); //
 	}
 
+	public static List<File> getProjectFilterFiles() {
+		if ( !resolveFilterPath().toFile().exists() ) {
+			ProjectConfiguration.getInstance().ensureProjectConfigurationInitialized();
```
diba1013: Can't we just assume, that the project is initialized, since a get() should not execute other stuff. I would suggest, just returning an empty list. 

It should be called at startup anyways. @martin-v

diba1013 (changes requested):

diba1013:
Also, please cleanup your commits

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -1,20 +1,34 @@
...
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import de.retest.recheck.configuration.ProjectConfiguration;
 
 public class SearchFilterFiles {
 
+	public static final String FILES_ENDING = ".filter";
```
diba1013:
```suggestion
	public static final String FILTER_EXTENSION = ".filter";
```

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -23,4 +37,14 @@ private SearchFilterFiles() {}
 				.collect( Collectors.toList() ); //
 	}
 
+	public static List<Path> getProjectFilterFiles() throws IOException {
+		if ( !resolveFilterPath().toFile().exists() ) {
```
diba1013: You could extract `resolveFilterPath()` into a variable, then you would not need a method for this.

diba1013 (changes requested):

diba1013 (approved):

PR #192, opened 2019-05-03
Feature/fix screenshot

enesoezel:


src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
```diff
@@ -2,8 +2,34 @@
...
+				(IdentifyingAttributesDifference) difference.getIdentifyingAttributesDifference();
+		for ( final AttributeDifference aDiff : diff.getAttributeDifferences() ) {
+			if ( aDiff.getKey().equals( OutlineAttribute.RELATIVE_OUTLINE ) ) {
+				return ((Rectangle) aDiff.getActual());
+			}
+		}
+		return getOutline( difference.getIdentifyingAttributes() );
+	}
+
+	public static Rectangle getActualAbsoluteOutline( final ElementDifference difference ) {
```
beatngu13, line 21: `getActualOutline` and this method are quite similar. Does it make sense to merge/reuse parts?

src/main/java/de/retest/recheck/ui/diff/ElementDifference.java
```diff
@@ -56,9 +56,9 @@ public Screenshot mark( final Screenshot screenshot, final Filter filter ) {
 		if ( childDifferences != null ) {
 			for ( final Difference childDifference : childDifferences ) {
 				for ( final ElementDifference compDiff : childDifference.getNonEmptyDifferences() ) {
-					if ( !filter.matches( element )
```
beatngu13, line 59: I think this diff doesn't belong in this commit?

src/main/java/de/retest/recheck/ui/descriptors/Attribute.java
```diff
@@ -10,7 +10,7 @@
 
 	private static final long serialVersionUID = 1L;
 
-	public static final double IGNORE_WEIGHT = 0.0d;
+	public static final double IGNORE_WEIGHT = Float.MIN_NORMAL;
```
beatngu13: Could you explain this change, commit message isn't very helpful? And isn't there also a `Double` variant?

beatngu13 (changes requested):

src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
```diff
@@ -9,24 +10,32 @@
 public class AttributeUtil {
 
 	public static Rectangle getActualOutline( final ElementDifference difference ) {
-		final IdentifyingAttributesDifference diff =
+		final IdentifyingAttributesDifference identifyingAttributesDifference =
 				(IdentifyingAttributesDifference) difference.getIdentifyingAttributesDifference();
-		for ( final AttributeDifference aDiff : diff.getAttributeDifferences() ) {
-			if ( aDiff.getKey().equals( OutlineAttribute.RELATIVE_OUTLINE ) ) {
-				return ((Rectangle) aDiff.getActual());
+		try {
```
diba1013: Why is there a `try catch`?

src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
```diff
@@ -1,9 +1,44 @@
...
+import de.retest.recheck.ui.diff.ElementDifference;
+import de.retest.recheck.ui.diff.IdentifyingAttributesDifference;
 
 public class AttributeUtil {
 
+	public static Rectangle getActualOutline( final ElementDifference difference ) {
+		final IdentifyingAttributesDifference identifyingAttributesDifference =
+				(IdentifyingAttributesDifference) difference.getIdentifyingAttributesDifference();
+		try {
+			if ( Objects.nonNull( identifyingAttributesDifference.getAttributeDifferences() ) ) {
```
diba1013: Just use `!= null` since there is no advantage to use `nonNull`. Also extract `identifyingAttributesDifference.getAttributeDifferences()` since you call it twice.

diba1013 (changes requested):

beatngu13:
Is this PR related to #195, which has been merged recently?

PR #193, opened 2019-05-03
Issue182: Load recheck.ignore file by suite

hansi-b:
This pull request pertains to the first half of https://github.com/retest/recheck-web/issues/182 : 

The RetestImpl now loads a second, optional, suite-specific filter in addition to the already loaded global one.

The new suite filter is loaded from the suite's golden master directory. It is loaded with the same mechanism as the global filter (e.g., also allows for a rules file). One difference is that a missing suite filter is not logged as an error (but as a debug message).

The remaining part of issue 182 is adding the step-wise filters (called level 3 in issue 182)

### DIFF
--- a/src/main/java/de/retest/recheck/RecheckImpl.java
+++ b/src/main/java/de/retest/recheck/RecheckImpl.java
@@ -62,7 +62,7 @@ public class RecheckImpl implements Recheck, SutStateLoader {
 
 	private final Map<String, DefaultValueFinder> usedFinders = new HashMap<>();
 	private final TestReplayResultPrinter printer;
-	private final Filter ignoreApplier;
+	private final Filter filter;
 
 	/**
 	 * Constructor that works purely with defaults. Default {@link FileNamerStrategy} assumes being called from within a
@@ -85,9 +85,9 @@ public class RecheckImpl implements Recheck, SutStateLoader {
 		final GlobalIgnoreApplier globalIgnoreApplier = RecheckIgnoreUtil.loadRecheckIgnore();
 		final GlobalIgnoreApplier suiteIgnoreApplier = RecheckIgnoreUtil.loadRecheckSuiteIgnore( getSuitePath() );
 
-		ignoreApplier = new CompoundFilter( Lists.newArrayList( globalIgnoreApplier, suiteIgnoreApplier ) );
+		filter = new CompoundFilter( Lists.newArrayList( globalIgnoreApplier, suiteIgnoreApplier ) );
 
-		printer = new TestReplayResultPrinter( usedFinders::get, ignoreApplier );
+		printer = new TestReplayResultPrinter( usedFinders::get, filter );
 	}
 
 	private static void ensureConfigurationInitialized() {
@@ -167,7 +167,7 @@ public class RecheckImpl implements Recheck, SutStateLoader {
 		suite.addTest( currentTestResult );
 		final TestReplayResult finishedTestResult = currentTestResult;
 		currentTestResult = null;
-		final Set<LeafDifference> uniqueDifferences = finishedTestResult.getDifferences( ignoreApplier );
+		final Set<LeafDifference> uniqueDifferences = finishedTestResult.getDifferences( filter );
 		logger.info( "Found {} not ignored differences in test {}.", uniqueDifferences.size(),
 				finishedTestResult.getName() );
 		if ( !uniqueDifferences.isEmpty() ) {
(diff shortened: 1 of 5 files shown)


src/main/java/de/retest/recheck/RecheckImpl.java
```diff
@@ -58,7 +62,7 @@
 
 	private final Map<String, DefaultValueFinder> usedFinders = new HashMap<>();
 	private final TestReplayResultPrinter printer;
-	private final GlobalIgnoreApplier ignoreApplier;
+	private final Filter ignoreApplier;
```
beatngu13: I think changing the field name would also be a good idea:

```suggestion
	private final Filter filter;
```

src/main/java/de/retest/recheck/RecheckImpl.java
```diff
@@ -77,7 +81,12 @@ public RecheckImpl( final RecheckOptions options ) {
 		fileNamerStrategy = options.getFileNamerStrategy();
 		suiteName = options.getSuiteName();
 		suite = SuiteReplayResultProvider.getInstance().getSuite( suiteName );
-		ignoreApplier = RecheckIgnoreUtil.loadRecheckIgnore();
+
+		final GlobalIgnoreApplier globalIgnoreApplier = RecheckIgnoreUtil.loadRecheckIgnore();
+		final GlobalIgnoreApplier suiteIgnoreApplier = RecheckIgnoreUtil.loadRecheckSuiteIgnore( getSuitePath() );
+
+		ignoreApplier = new CompoundFilter( Lists.newArrayList( globalIgnoreApplier, suiteIgnoreApplier ) );
```
beatngu13: IMHO it is good to not rely on Guava if there is similar JDK method:

```suggestion
		ignoreApplier = new CompoundFilter( Arrays.asList( globalIgnoreApplier, suiteIgnoreApplier ) );
```

src/main/java/de/retest/recheck/review/workers/LoadFilterWorker.java
```diff
@@ -18,13 +20,30 @@
 public class LoadFilterWorker {
 
 	private final Counter counter;
+	private final Optional<Path> ignoreFile;
```
beatngu13: I know that this is a controversial topic, but I wouldn't use `Optional` in fields.

See Stuart Marks' (who works for the core libraries team in the JDK group at Oracle) talk "Optional – The Mother of All Bikesheds" from Devoxx 2016 at [54:04](https://www.youtube.com/watch?v=Ej0sss6cq14&feature=youtu.be&t=54m04s):

>Why Not Use Optional in Fields?
>
>* More a style issue than a correctness issue
>   * usually there's a better way to model absence of a value
>   * use of Optional in fields often arises from slavish desire to eliminate nullable fields
>   * remember, eliminating nulls isn't a goal of Optional
>* Using Optional in fields…
>   * creates another object for every field
>   * introduces a dependent load from memory on every field read
>   * clutters up your code
>   * to what benefit? ability to chain methods?

src/test/java/de/retest/recheck/RecheckImplTest.java
```diff
@@ -51,7 +41,8 @@ void using_strange_stepText_should_be_normalized() throws Exception {
 		} catch ( final Exception e ) {
 			// ignore exception, fear AssertionErrors...
 		}
-		assertThat( check.wasCalled ).isTrue();
+		Mockito.verify( check ).createFileNamer( "de.retest.recheck.RecheckImplTest" );
```
beatngu13: We almost always use static importants for this reduce noise.

src/test/java/de/retest/recheck/review/workers/LoadFilterWorkerTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import de.retest.recheck.configuration.ProjectConfiguration;
+import de.retest.recheck.review.counter.NopCounter;
+
+public class LoadFilterWorkerTest {
+
+	@Test
+	void loading_without_ignore_file_should_fail( @TempDir final Path root ) throws Exception {
+		final LoadFilterWorker worker = new LoadFilterWorker( NopCounter.getInstance(), root.toFile() );
+		assertThrows( IllegalArgumentException.class, () -> worker.load() );
```
beatngu13: We usually use AssertJ for assertions as they are—in our opinion—more flexible and readable.

beatngu13 (changes requested):

beatngu13 (approved):

hansi-b:
Sorry, only got around to this now - thanks for the adaptations and the link to the interesting talk!

PR #194, opened 2019-05-06
Change file to path

alig01:


src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -25,11 +26,11 @@
 
 	private SearchFilterFiles() {}
 
-	public static List<File> getDefaultFilterFiles() {
+	public static List<Path> getDefaultFilterFiles() {
 		return defaultWebFilter.stream() //
 				.map( filter -> SearchFilterFiles.class.getResource( filter ) ) //
 				.filter( Objects::nonNull ) //
-				.map( resource -> new File( resource.getFile() ) ) //
+				.map( resource -> Paths.get( URI.create( resource.toString() ) ) ) //
```
beatngu13, line 33: I think this should also work:

```suggestion
				.map( url -> Paths.get( url.toURI() ) ) //
```

beatngu13 (changes requested):

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
alig01: it would lead to an ugly try catch block
```suggestion
.map( url -> {
	try {
		return Paths.get( url.toURI() );
		} catch ( final URISyntaxException e ) {
		}
		return null;
		} )
```

beatngu13 (approved):

PR #196, opened 2019-05-06
Feature/ret 1642 get list all filter

alig01:


src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -26,6 +27,21 @@
...
+		final List<Path> filterFiles = new ArrayList<>();
+		if ( !getDefaultFilterFiles().isEmpty() ) {
+			filterFiles.addAll( getDefaultFilterFiles() );
+		}
+		try {
+			if ( !getDefaultFilterFiles().isEmpty() ) {
+				filterFiles.addAll( getProjectFilterFiles() );
+			}
+		} catch ( final IOException exc ) {
+			exc.printStackTrace();
```
beatngu13: As suggested by Sonar: Please fix the exception handling.

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -26,6 +27,21 @@
 
 	private SearchFilterFiles() {}
 
+	public static List<Path> getAllFilterFiles() {
+		final List<Path> filterFiles = new ArrayList<>();
+		if ( !getDefaultFilterFiles().isEmpty() ) {
```
beatngu13: What is the difference between this line and line 36? And why is 36 wrapped inside a try-catch?

src/test/java/de/retest/recheck/ignore/SearchFilterFilesTest.java
```diff
@@ -18,11 +18,18 @@
 
 class SearchFilterFilesTest {
 
+	@Test
+	void getAllFilterFiles_should_find_all_filters() {
+		final List<Path> allFilterFiles = SearchFilterFiles.getAllFilterFiles();
+		assertThat( allFilterFiles.stream().map( path -> path.getFileName().toString() ) )
+				.contains( "positioning.filter", "visibility.filter" );
```
beatngu13: I think you should add a user filter file (?) and test this, because it looks like this is currently the same test case as `getAllFilterFiles_should_find_all_filters()`.

beatngu13 (changes requested):

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
alig01: getProjectFilterFiles() using Files.walk which can cause an IOException

beatngu13 (approved):

PR #198, opened 2019-05-06
We need to be able to set the retestId, or else the log output is wrong

roesslerj:
Ok ... finally there doesn't seem to be a way around it. We need the returned Element have the same retestId as the old element...

### DIFF
--- a/src/main/java/de/retest/recheck/ui/descriptors/Element.java
+++ b/src/main/java/de/retest/recheck/ui/descriptors/Element.java
@@ -108,8 +108,8 @@ public class Element implements Serializable, Comparable<Element> {
 	 * Note that the retestId is immutable, so this method returns a new Element that is a copy of the old one with the
 	 * new retestId.
 	 */
-	public Element setRetestId( final String retestId ) {
-		return Element.create( retestId, this, identifyingAttributes, attributes );
+	public Element applyRetestId( final String retestId ) {
+		return Element.create( retestId, parent, identifyingAttributes, attributes );
 	}
 
 	protected List<Element> createNewElementList( final ActionChangeSet actionChangeSet,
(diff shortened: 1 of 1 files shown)


src/main/java/de/retest/recheck/ui/descriptors/Element.java
```diff
@@ -104,6 +104,14 @@ public Element applyChanges( final ActionChangeSet actionChangeSet ) {
 		return element;
 	}
 
+	/**
+	 * Note that the retestId is immutable, so this method returns a new Element that is a copy of the old one with the
+	 * new retestId.
+	 */
+	public Element setRetestId( final String retestId ) {
```
martin-v: I wouldn't call the method `set` because it don't work like a typical `set*` method. I prefer something like `applyRetestId`.

src/main/java/de/retest/recheck/ui/descriptors/Element.java
```diff
@@ -104,6 +104,14 @@ public Element applyChanges( final ActionChangeSet actionChangeSet ) {
...
+		return Element.create( retestId, this, identifyingAttributes, attributes );
```
martin-v: In most cases `this` isn't `parent` ;)

martin-v (changes requested):

src/main/java/de/retest/recheck/ui/descriptors/Element.java
roesslerj: Oh, wow. This looks like a major error: I copied this code from 5 lines above (`public Element applyChanges( final ActionChangeSet actionChangeSet ) {`). Someone needs to look into that.

martin-v (approved):

PR #199, opened 2019-05-06
Adding the RetestIdProvider to the options...

roesslerj:
... which is more consistent than a system property.

### DIFF
--- a/src/main/java/de/retest/recheck/util/RetestIdProviderUtil.java
+++ b/src/main/java/de/retest/recheck/util/RetestIdProviderUtil.java
@@ -43,6 +43,10 @@ public class RetestIdProviderUtil {
 		return getRetestId( getConfiguredRetestIdProvider(), identifyingAttributes );
 	}
 
+	public static void setRetestIdProvider( final RetestIdProvider retestIdProvider ) {
+		configured = retestIdProvider;
+	}
+
 	/**
 	 * Testing only.
 	 *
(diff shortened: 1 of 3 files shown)


src/main/java/de/retest/recheck/util/RetestIdProviderUtil.java
```diff
@@ -43,6 +43,10 @@ public static String getRetestId( final IdentifyingAttributes identifyingAttribu
 		return getRetestId( getConfiguredRetestIdProvider(), identifyingAttributes );
 	}
 
+	public static void setRetestIdProvider( final RetestIdProvider retestIdProvider ) {
```
beatngu13, line 46: Having a static setter is evil. 💩 We also already have `RetestIdProvider.RETEST_ID_PROVIDER_PROPERTY`?

roesslerj: Right. So what is your proposed solution?

beatngu13: In recheck-web, we have recently introduced the following change:

https://github.com/retest/recheck-web/commit/115d2a9fdbd2b628e104708e8191255220aa488f#diff-307694955239e6ca83167b42108d04fb

Now, e.g., the `RecheckSeleniumAdapter` holds a reference to an instance of `RetestIdProvider`. This allows to reuse the adapter in surili.

This may be problem with this change here. If the ID provider type changes during runtime (via the setter), other classes might use an obsolete implementation—depending on the lifecycle of the class, I'm not sure.

Regarding the setter itself:

* Do we need both, the property and the options?
* Maybe a [dynamic factory](https://medium.com/@mhd.durrah/factory-pattern-the-dynamic-way-with-java-8-3ca5ab48a9cf) is better suited?

beatngu13:
Will close this until we have worked it out. ✌️

PR #200, opened 2019-05-06
Feature/add alignment threshold

roesslerj:


src/test/java/de/retest/recheck/ui/diff/ElementDifferenceFinderTest.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.Test;
+
+import de.retest.recheck.ui.descriptors.Element;
+
+public class ElementDifferenceFinderTest {
+
+	@Test
```
beatngu13, line 17: New tests should use JUnit 5, then we can also omit `public` both on class and method level.

src/test/java/de/retest/recheck/ui/diff/AlignmentTest.java
```diff
@@ -0,0 +1,153 @@
...
+				new Attributes() );
+		final Map<Element, Element> mapping = Alignment.toMapping( Collections.singletonList( expected ) );
+
+		final Element actual = mapping.get( expectedClone );
+
+		assertThat( actual ).isSameAs( expected );
+	}
+
+	@Test
+	public void simple_one_on_one_alignment() throws Exception {
```
beatngu13: Omit `public`.

src/test/java/de/retest/recheck/ui/diff/AlignmentTest.java
```diff
@@ -144,6 +154,51 @@ public void alignment_with_less_parents() throws Exception {
...
+		attributes.put( "shown", false );
+
+		final Element actual = Element.create( "otherRetestId", actualRoot,
+				new IdentifyingAttributes( actualIdentifying ), actualAttributes.immutable() );
+		actualRoot.addChildren( actual );
+
+		final Alignment alignment = Alignment.createAlignment( root, actualRoot );
+		final Element mappedActual = alignment.get( expected );
+
+		Assertions.assertThat( mappedActual ).isNotEqualTo( actual );
```
beatngu13: Use static imports.

src/test/java/de/retest/recheck/ui/diff/AlignmentTest.java
```diff
@@ -144,6 +154,51 @@ public void alignment_with_less_parents() throws Exception {
...
+
+		final Element expected =
+				Element.create( "retestId", root, new IdentifyingAttributes( identifying ), attributes.immutable() );
+		root.addChildren( expected );
+
+		final Element actualRoot = buildEqual( "html[1]", Root.class );
+
+		final List<Attribute> actualIdentifying = new ArrayList<>();
+		identifying.add( new StringAttribute( "text",
+				"file:///Users/roessler/Documents/startup/workspace/recheck-web/src/test/resources/pages/non-existent.html" ) );
```
beatngu13, line 180: Why this huge local path as attribute value? Wouldn't it be sufficient to use something simpler that differs from `"change-for-div"`?

src/main/java/de/retest/recheck/Properties.java
```diff
@@ -11,6 +11,9 @@
 	public static final double WINDOW_MATCH_THRESHOLD = 0.8;
 	public static final double WINDOW_CHILDS_MATCH_THRESHOLD = 0.5;
 
+	public static final double ELEMENT_MATCH_THRESHOLD = 0.5;
```
beatngu13:
```suggestion
	public static final double ELEMENT_MATCH_THRESHOLD_DEFAULT = 0.5;
```

src/main/java/de/retest/recheck/ui/diff/Alignment.java
```diff
@@ -147,12 +147,30 @@ public static Alignment createAlignment( final Element expectedComponent, final
...
 			matches.put( bestMatch.element, new Match( bestMatch.similarity, expected ) );
 		}
 		return alignment;
 	}
 
+	private static Double getRequiredMinimumMatch() {
+		final String value = System.getProperty( Properties.ELEMENT_MATCH_THRESHOLD_PROPERTY );
+		try {
+			return Double.parseDouble( value );
+		} catch ( final Exception e ) {
```
beatngu13: Check `value` for `null`, then we don't have to catch the generic `Exception` type but only `NumberFormatException`.

beatngu13 (changes requested):

src/test/java/de/retest/recheck/ui/diff/ElementDifferenceFinderTest.java
roesslerj: Yes, but this is a copy&paste test from retest...

src/test/java/de/retest/recheck/ui/diff/AlignmentTest.java
roesslerj: As of TDD, this is an exact replication of the problem. Why is that a problem?

beatngu13: This is not a big deal, but it is noise IMHO. When we look at this in a few months, we will probably be confused.

The cause for the problem is not the huge local path. The test case name says "very different elements should not be aligned", so I assume we also don't want "foo" and "barbaz4711" to be aligned? Generalizing the bug may help us to get a better understanding.

It is up to you, we can also keep it as it is.

beatngu13 (approved):

beatngu13:
Done, but could you please take care of the SonarCloud issues? I think you can probably mark a lot of them as "won't resolve". Let me know when you need help.

PR #210, opened 2019-05-15
Ignored identifying attributes should cause no difference

beatngu13:
Commit ae83da9 makes sure that the weight is considered during diffing. Since we have changed this from 0 to `Double.MIN_NORMAL`, I assume we don't want absolute outline differences anymore? What is the expected behavior for `AttributeUtilTest#getActualAbsoluteOutline_should_return_actual_absolute_outline()`? There is no difference no more.

### DIFF
--- a/src/main/java/de/retest/recheck/ui/diff/IdentifyingAttributesDifferenceFinder.java
+++ b/src/main/java/de/retest/recheck/ui/diff/IdentifyingAttributesDifferenceFinder.java
@@ -26,8 +26,7 @@ public class IdentifyingAttributesDifferenceFinder {
 		final List<AttributeDifference> attributeDifferences = new ArrayList<>();
 
 		for ( final Attribute expectedAttr : expected.getAttributes() ) {
-			if ( Double.compare( expectedAttr.getWeight(), Attribute.IGNORE_WEIGHT ) == 0
-					|| expectedAttr.isNotVisible() ) {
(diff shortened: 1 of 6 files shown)


src/main/java/de/retest/recheck/ui/diff/IdentifyingAttributesDifferenceFinder.java
```diff
@@ -15,19 +15,25 @@
...
+	IdentifyingAttributesDifference differenceFor( final IdentifyingAttributes expected,
+			final IdentifyingAttributes actual, final GloballyIgnoredAttributes ignored ) {
 		Objects.requireNonNull( expected, "Expected cannot be null!" );
 		Objects.requireNonNull( actual, "Actual cannot be null!" );
 
 		final List<AttributeDifference> attributeDifferences = new ArrayList<>();
 
 		for ( final Attribute expectedAttr : expected.getAttributes() ) {
-			if ( Double.compare( expectedAttr.getWeight(), 0.0d ) == 0 || expectedAttr.isNotVisible() ) {
+			if ( Double.compare( expectedAttr.getWeight(), Attribute.IGNORE_WEIGHT ) == 0
```
beatngu13: `if` should be removed, because weight is supposed to be only used for matching, not for differencing. (Adapt tests accordingly.)

roesslerj (approved):

PR #226, opened 2019-05-19
Feature/improve error handling

roesslerj:


src/main/java/de/retest/recheck/ignore/RecheckIgnoreUtil.java
```diff
@@ -36,20 +37,24 @@ public static GlobalIgnoreApplier loadRecheckIgnore() {
 		try {
 			final LoadFilterWorker loadFilterWorker = new LoadFilterWorker( NopCounter.getInstance() );
 			return loadFilterWorker.load();
+		} catch ( final FileNotFoundException e ) {
+			logger.error( "Could not find recheck ignore file." );
```
beatngu13, line 41: We should also log the actual exception (like below):

```suggestion
			logger.error( "Could not find recheck ignore file.", e );
```

src/main/java/de/retest/recheck/ignore/RecheckIgnoreUtil.java
```diff
@@ -36,20 +37,24 @@ public static GlobalIgnoreApplier loadRecheckIgnore() {
...
+		return GlobalIgnoreApplier.create( NopCounter.getInstance() );
 	}
 
 	public static GlobalIgnoreApplier loadRecheckSuiteIgnore( final File ignoreFilesBasePath ) {
 		try {
 			final LoadFilterWorker loadFilterWorker =
 					new LoadFilterWorker( NopCounter.getInstance(), ignoreFilesBasePath );
 			return loadFilterWorker.load();
+		} catch ( final FileNotFoundException e ) {
+			logger.debug( "Ignoring missing suite ignore file." );
```
beatngu13, line 54: We should also log the actual exception (like below):

```suggestion
			logger.debug( "Ignoring missing suite ignore file.", e );
```

beatngu13 (changes requested):

src/main/java/de/retest/recheck/ignore/RecheckIgnoreUtil.java
roesslerj: Actually, when you have a look at the log, (e.g. here https://travis-ci.com/retest/recheck-web/builds/112459315), not logging that exception was one of the motivations for that change. It adds a lot of noice...

src/main/java/de/retest/recheck/ignore/RecheckIgnoreUtil.java
roesslerj: Same here.

PR #239, opened 2019-05-21
Fix loading of filters from jar

diba1013:
since the ZipFileSystemProvider does not create FileSystems as necessary, we have to create our own but then are forced to directly use it as it must be closed afterwards.


diba1013:
Wrong target branch, creating new branch for release

PR #240, opened 2019-05-21
Fix loading of filters from jar

diba1013:
since the ZipFileSystemProvider does not create FileSystems as necessary, we have to create our own but then are forced to directly use it as it must be closed afterwards.

### DIFF
--- a/src/main/java/de/retest/recheck/ignore/FilterLoader.java
+++ b/src/main/java/de/retest/recheck/ignore/FilterLoader.java
@@ -1,0 +1,21 @@
+package de.retest.recheck.ignore;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+public interface FilterLoader {
+
+	Filter load() throws IOException;
+
+	static FilterLoader load( final Path path ) {
+		return () -> Filters.load( path );
+	}
+
+	static FilterLoader provide( final Path path ) throws IOException {
+		return provide( Filters.load( path ) );
+	}
+
+	static FilterLoader provide( final Filter filter ) {
+		return () -> filter;
+	}
+}
--- a/src/main/java/de/retest/recheck/ignore/Filters.java
+++ b/src/main/java/de/retest/recheck/ignore/Filters.java
@@ -1,0 +1,21 @@
+package de.retest.recheck.ignore;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import de.retest.recheck.review.ignore.io.Loaders;
+
+public class Filters {
+
+	public static Filter load( final Path path ) throws IOException {
+		try ( final Stream<String> filterFileLines = Files.lines( path ) ) {
+			return Loaders.load( filterFileLines ) //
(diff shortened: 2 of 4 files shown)


src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -32,50 +35,64 @@
...
-					.collect( Collectors.collectingAndThen( Collectors.toList(), CompoundFilter::new ) );
-		} catch ( final IOException e ) {
-			throw new UncheckedIOException( "Could not load '" + filter + "' file.", e );
-		}
+				.collect( Collectors.toMap( Pair::getLeft, pair -> {
+					final FilterLoader loader = pair.getRight();
+					try {
+						return loader.load();
+					} catch ( final IOException e ) {
+						log.error( "Could not load Filter for '{}'.", pair.getLeft() );
```
beatngu13, line 93: I would suggest to ignore the actual exception:

```suggestion
						log.error( "Could not load Filter for '{}'.", pair.getLeft(), e );
```

src/main/java/de/retest/recheck/ignore/FilterLoader.java
```diff
@@ -0,0 +1,21 @@
...
+
+	static FilterLoader load( final Path path ) {
+		return () -> Filters.load( path );
+	}
+
+	static FilterLoader provide( final Path path ) throws IOException {
+		return provide( Filters.load( path ) );
+	}
+
+	static FilterLoader provide( final Filter filter ) {
```
beatngu13, line 18: What is this for? Isn't this like `identitiy()`?

src/main/java/de/retest/recheck/ignore/Filters.java
```diff
@@ -0,0 +1,21 @@
...
+
+import de.retest.recheck.review.ignore.io.Loaders;
+
+public class Filters {
+
+	public static Filter load( final Path path ) throws IOException {
+		try ( final Stream<String> filterFileLines = Files.lines( path ) ) {
+			return Loaders.load( filterFileLines ) //
+					.filter( Filter.class::isInstance ) //
+					.map( Filter.class::cast )//
```
beatngu13, line 17: Missing whitespace:

```suggestion
					.map( Filter.class::cast ) //
```

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -32,50 +35,64 @@
...
-				.collect( Collectors.toList() ); //
+				.collect( Collectors.toList() );
+	}
+
+	private static Pair<Path, FilterLoader> loadFilterFromUri( final URI uri ) {
+		try {
+			final Path path = Paths.get( uri );
+			return Pair.of( path, FilterLoader.load( path ) );
+		} catch ( final FileSystemNotFoundException e ) {
+			try ( final FileSystem fs = FileSystems.newFileSystem( uri, Collections.emptyMap() ) ) {
```
beatngu13, line 59: Nested try-catch is quite ugly, maybe extract a descriptive method?

beatngu13 (changes requested):

roesslerj (approved):

PR #242, opened 2019-05-21
Don't throw an exception when the project could not be found

diba1013:
instead just return an empty list

### DIFF
--- a/src/main/java/de/retest/recheck/configuration/ProjectConfiguration.java
+++ b/src/main/java/de/retest/recheck/configuration/ProjectConfiguration.java
@@ -4,6 +4,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Optional;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


src/main/java/de/retest/recheck/configuration/ProjectConfiguration.java
```diff
@@ -40,10 +40,12 @@ public static ProjectConfiguration getInstance() {
 		return instance;
 	}
 
+	public Optional<Path> getProjectConfigFolder() {
```
beatngu13, line 43: Find it hard to distinguish between `getProjectConfigFolder()` and `findProjectConfigFolder()`, only the return types differ.

As the latter potentially throws an exception (like `Optional#get()`) and the former doesn't, how about:

* `getProjectConfigFolder()` => `searchProjectConfigFolder()` returns `Optional`.
* `findProjectConfigFolder()` => `getProjectConfigFolder()` potentially throws an exception.

I'm not sure, ideas?

diba1013: I would have removed it, or at least made it private. But I could not be bothered to look at where it is used...

diba1013: The `get...` is related to `ProjectRootFinderUtil.getProjectRoot()` where also an `Optional` is returned.

beatngu13: OK, I see, makes sense!

beatngu13 (approved):

PR #243, opened 2019-05-21
Do not save paths for filters

diba1013:
to enable overwriting of project specific filters since they have the different paths from the jar and the os file system.

**Please feel free to write a test, I have not the time now and leave it as: It should work!?**

### DIFF
--- a/src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
+++ b/src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
@@ -90,17 +90,25 @@ public class SearchFilterFiles {
 		}
 	}
 
-	public static Map<String, Filter> toFileNameFilterMapping( final List<Pair<String, FilterLoader>> paths ) {
-		return paths.stream() //
-				.collect( Collectors.toMap( Pair::getLeft, pair -> {
-					final FilterLoader loader = pair.getRight();
-					try {
-						return loader.load();
-					} catch ( final IOException e ) {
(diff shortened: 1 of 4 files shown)


src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -35,12 +35,12 @@
 
 	private SearchFilterFiles() {}
 
-	public static List<Pair<Path, FilterLoader>> getAllFilterFiles() {
+	public static List<Pair<String, FilterLoader>> getAllFilterFiles() {
```
beatngu13: Wasn't it your suggestion to use `Path` instead of `String` to reflect that filters don't have a name but only a path? Moreover, `Path#getFileName()` only returns the filename, so doesn't this cause duplicates if the user creates a filter with the same name as a predefined one?

beatngu13:
Wasn't it your suggestion to use `Path` instead of `String` to reflect that filters don't have a name but only a path?

Moreover, `Path#getFileName()` only returns the filename, so doesn't this cause duplicates if the user creates a filter with the same name as a predefined one? If we want the user to allow overwritting predefined filters, we have to implement this in `toPathFilterMapping( … )` (which now should be named `toFilenameFilterMapping( … )` to prioritize user-defined filters.

As I think this is not crucial for testing, I will keep this open and release a beta anyway. We can still merge this afterwards.

beatngu13:
@diba1013 adapted the implementation as discussed, would mind having a look at it?

beatngu13 (approved):

PR #259, opened 2019-05-24
Feature/improve filter

roesslerj:


src/main/resources/filter/web/positioning.filter
```diff
@@ -1,21 +1,27 @@
...
+attribute=clip
+
 attribute=margin-top
 attribute=margin-right
 attribute=margin-bottom
 attribute=margin-left
+
 attribute=object-position
 attribute=outline
+attribute=absolute-outline
```
diba1013, line 19: If your fixes were successful, `absolute-outline` should never pop up.

diba1013 (changes requested):
I did not catch, when those new filters were added, but I think, that filter names should be as simple as possible: `invisible` or rather `invisibility` since `visibility` - `style`, ...? How about invisible html elements like `script`, `meta`, `use`, ... - do we catch these?

You cannot just add new filter files. If you want them to be usable, add them to `SearchFilterFiles`!

Also, we really someday need a way to reference filters: `!file: path/to/file` or `!filter: name`

src/main/resources/filter/web/positioning.filter
roesslerj: That's not true. We changed that...

roesslerj:
Screen Shot 2019-05-24 at 24 05 19  14 01 53
"visibility" turned out to be not a good name, as it doesn't convey what is meant. It's not clear to the user what will happen when he activates that filter.

diba1013 (approved):

PR #260, opened 2019-05-27
Feature/ret 1604 adjust apply changes bar

kema91:
RET-1604 Apply changes is visible when it should be hidden.

### DIFF
--- a/src/main/java/de/retest/recheck/review/GlobalIgnoreApplier.java
+++ b/src/main/java/de/retest/recheck/review/GlobalIgnoreApplier.java
@@ -57,9 +57,13 @@ public class GlobalIgnoreApplier implements Filter {
 		remove( filter -> filter.matches( element ) );
 	}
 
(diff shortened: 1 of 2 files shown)


src/main/java/de/retest/recheck/review/GlobalIgnoreApplier.java
```diff
@@ -57,11 +57,15 @@ public void unignoreElement( final Element element ) {
 		remove( filter -> filter.matches( element ) );
 	}
 
-	public void add( final Filter filter ) {
+	private void add( final Filter filter ) {
 		filtered.add( filter );
 		counter.add();
 	}
 
+	public void addJSRuleFile( final Filter filter ) {
```
roesslerj: Rename z.B. `addWithoutCounting` oder so...

roesslerj (changes requested):

roesslerj (approved):

PR #267, opened 2019-05-31
Load ignore files/rules per step

hansi-b:
This pull request proposes a solution for part 2 (level 3) of https://github.com/retest/recheck-web/issues/182: Having an optional filter for each test step.

The two main changes are in `RecheckImpl`:

1.  It gets a new stateful field `File currentStepDir` (kept in a fashion similar to the already existing `currentTestResult`). This makes the information about the current step available in `capTest` to construct a step-specific filter from the suite filter.
1. The former `TestReplayResultPrinter` field is now locally constructed in `RecheckImpl::getDifferencesErrorMessage` using the step-specific filter.

The loading of the step-specific filter delegates to the same logic used by the global and suite filters in `RecheckIgnoreUtil`. As with suite filters, a missing step filter will only log a debug message.

---
On a related note, the `GlobalIgnoreApplier` should perhaps be renamed: It is now used for 1. the global filter, 2. the suite filter, and 3. the step filter. It is rather like a mutable version of the `CompoundFilter`; it also happens to be the type used for loading and saving. So it seems to be something like a `MutablePersistableCompoundFilter`. IMHO, its inner class `PersistableGlobalIgnoreApplier` could also be replaced with a simpler list copy during (de)serialization.

### DIFF
--- a/src/main/java/de/retest/recheck/RecheckImpl.java
+++ b/src/main/java/de/retest/recheck/RecheckImpl.java
@@ -59,9 +59,9 @@ public class RecheckImpl implements Recheck, SutStateLoader {
 	private final Filter globalAndSuiteFilter;
 
 	private final Map<String, DefaultValueFinder> usedFinders = new HashMap<>();
+	private final Map<ActionReplayResult, Filter> effectiveFilterByReplayResult = new HashMap<>();;
 
 	private TestReplayResult currentTestResult;
-	private File currentStepDir;
 
 	/**
 	 * Constructor that works purely with defaults. Default {@link FileNamerStrategy} assumes being called from within a
@@ -125,7 +125,7 @@ public class RecheckImpl implements Recheck, SutStateLoader {
 		usedFinders.put( currentStep, defaultFinder );
 
 		final FileNamer fileNamer = createFileName( currentStep );
-		currentStepDir = fileNamer.getFile( Properties.GOLDEN_MASTER_FILE_EXTENSION );
+		final File currentStepDir = fileNamer.getFile( Properties.GOLDEN_MASTER_FILE_EXTENSION );
 
 		final SutState actual = RecheckSutState.convert( toVerify, adapter );
 		final SutState expected = loadExpected( currentStepDir );
@@ -136,7 +136,13 @@ public class RecheckImpl implements Recheck, SutStateLoader {
 		final RecheckDifferenceFinder finder =
 				new RecheckDifferenceFinder( adapter.getDefaultValueFinder(), currentStep, currentStepDir.getPath() );
 
-		return finder.findDifferences( actual, expected );
+		final ActionReplayResult actionResult = finder.findDifferences( actual, expected );
+
+		final GlobalIgnoreApplier stepIgnoreApplier = RecheckIgnoreUtil.loadRecheckIgnore( currentStepDir );
+		effectiveFilterByReplayResult.put( actionResult,
+				new CompoundFilter( Arrays.asList( globalAndSuiteFilter, stepIgnoreApplier ) ) );
+
+		return actionResult;
 	}
 
 	private File getSuitePath() {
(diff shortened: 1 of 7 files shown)


beatngu13:
Thanks a lot for the PR! 🙏 We will go through this ASAP.

src/main/java/de/retest/recheck/RecheckImpl.java
```diff
@@ -165,13 +163,18 @@ public void capTest() {
 		suite.addTest( currentTestResult );
 		final TestReplayResult finishedTestResult = currentTestResult;
 		currentTestResult = null;
-		final Set<LeafDifference> uniqueDifferences = finishedTestResult.getDifferences( filter );
+
+		final GlobalIgnoreApplier stepIgnoreApplier = RecheckIgnoreUtil.loadRecheckIgnore( currentStepDir );
+		final Filter stepFilter = new CompoundFilter( Arrays.asList( suiteFilter, stepIgnoreApplier ) );
```
beatngu13: Wouldn't this need 3 filters?

1. Global `recheck.ignore`
2. Suite `recheck.ignore`
3. Step `recheck.ignore`

beatngu13 (changes requested):
Looks good, just a few adjustments. 👍 Thinking about how a possible test could look like, but this requires a lot of I/O mocking I guess…

src/main/java/de/retest/recheck/RecheckImpl.java
hansi-b: That's a good point - the three filters are already there, but in an obfuscated way: The global one is hiding in the misleadingly named `suiteFilter`, which is already constructed as a compound filter from global and suite filters in the constructor.
I guess we could either rename it to `globalAndSuiteFilter` (or similar) for clarification, or we can keep both filters in the `RecheckImpl` and then have the three item list you had expected. Any preference?

src/main/java/de/retest/recheck/ignore/RecheckIgnoreUtil.java
```diff
@@ -45,15 +45,18 @@ public static GlobalIgnoreApplier loadRecheckIgnore() {
 		return GlobalIgnoreApplier.create( NopCounter.getInstance() );
 	}
 
-	public static GlobalIgnoreApplier loadRecheckSuiteIgnore( final File ignoreFilesBasePath ) {
+	/**
+	 * Loads a filter from the argument base path, e.g., test suite and step filters.
+	 */
+	public static GlobalIgnoreApplier loadRecheckIgnore( final File ignoreFilesBasePath ) {
```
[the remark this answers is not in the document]
hansi-b: Yes, in the end it was mainly the parallelism with `loadGlobalRecheckIgnore`: From the point of view of `RecheckImpl`, both methods could just return a `Filter`, but I suspect the specific return type for the global call matters in the non-public code base. As a suggestion, I have checked in a more specific javadoc.

src/main/java/de/retest/recheck/ignore/RecheckIgnoreUtil.java
```diff
@@ -45,15 +45,18 @@ public static GlobalIgnoreApplier loadRecheckIgnore() {
 		return GlobalIgnoreApplier.create( NopCounter.getInstance() );
 	}
 
-	public static GlobalIgnoreApplier loadRecheckSuiteIgnore( final File ignoreFilesBasePath ) {
+	/**
+	 * Loads a filter from the argument base path, e.g., test suite and step filters.
```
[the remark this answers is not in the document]
hansi-b: Thanks, added that with the commit for the other javadoc change below.

beatngu13 (changes requested):
Please rename the field and then I think we are good to go. In the meantime, I will try to come up with a test case (feel free to propose one). And if you want to, you can add yourself to the `<contributors>` section of the `pom.xml`.

src/main/java/de/retest/recheck/RecheckImpl.java
hansi-b: Oh my! Good thing you asked about a test: It made me recheck my manual HTML test example to see if I could extract something unit-like. Instead, after a few trials my palm met my forehead: The code has an elementary conceptual problem (which is obvious in hindsight, don't know how I missed it then):

Currently, it only loads one step filter in `RecheckImpl::capTest`, i.e., only for the last step in a test. In general, `RecheckImpl` does step-wise stuff _mostly_ via the `ActionReplayResult`s, but `capTest` can currently only use one suite+global filter when finalizing the test results.

Unfortunately, I see no elegant way to fix this: Both `TestReplayResult::getDifferences` and the whole `TestReplayResultPrinter` currently also assume a single, constant filter when accumulating results from different `ActionReplayResult`s. One alternative would be for them to use a `Map<ActionReplayResult, Filter>` to look up the (step-specific) filter per action; but this seems a little hackish to me, and it would change more public signatures and associated behaviour than this feature might be worth. I have checked these changes in so you can take a look yourself.

I have also verified the changed code with my little manual html test (which Github won't let me attach here, strangely), and I am very certain that it now works as desired, but I would understand completely if you think it might be better to close this pull request for the moment. (The suite filters, part 1 of this issue, are not affected.)

beatngu13:
First of all: Sorry for the late reply!

Glad you find that issue—it is indeed not straightforward to come up with a proper solution. And as you said before, `Map<ActionReplayResult, Filter>` seems a bit hacky.

I would suggest to keep this PR open, we will discuss possible alternatives in https://github.com/retest/recheck-web/issues/182. But in any case, we need tests for this (although we should probably wait until we know which way we want to go).

We will keep you updated, and thanks again for your contribution!

diba1013:
We decided that we currently do not have the resources to properly handle this PR (as probably stated due to the long time this PR was opened). However, we will revise this once there is need.

Again, thanks for the contribution and sorry for the long delay.

PR #270, opened 2019-06-04
Feature/improve logging

roesslerj:
After approving the other branch...

### DIFF
--- a/src/main/java/de/retest/recheck/review/ignore/io/ErrorHandlingLoader.java
+++ b/src/main/java/de/retest/recheck/review/ignore/io/ErrorHandlingLoader.java
@@ -3,7 +3,9 @@ package de.retest.recheck.review.ignore.io;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public final class ErrorHandlingLoader implements Loader<Void> {
+import de.retest.recheck.review.ignore.FilterPreserveLineLoader.FilterPreserveLine;
+
+public final class ErrorHandlingLoader implements Loader<FilterPreserveLine> {
 
 	private static final Logger logger = LoggerFactory.getLogger( ErrorHandlingLoader.class );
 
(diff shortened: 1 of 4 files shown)


src/main/java/de/retest/recheck/review/ignore/io/ErrorHandlingLoader.java
```diff
@@ -0,0 +1,33 @@
+package de.retest.recheck.review.ignore.io;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import de.retest.recheck.review.ignore.FilterPreserveLineLoader.FilterPreserveLine;
+
+public final class ErrorHandlingLoader implements Loader<FilterPreserveLine> {
```
diba1013, line 8: `FilterPreserveLine` is a `Filter`, right? If so, please use `Filter`.

src/test/resources/de/retest/recheck/review/workers/recheck.ignore
```diff
@@ -6,6 +6,7 @@ matcher: retestid=banner, attribute-regex: .*color
 matcher: xpath=/html[1]/div[1]/div[1]/div[1]
 matcher: xpath=/html[1]/div[1]/div[1]/div[2], attribute: background-url
 matcher: type=div, attribute-regex: .*color
+attribute:outline
```
diba1013: Maybe add a comment, that this is intended to be an error.

diba1013 (changes requested):

src/main/java/de/retest/recheck/review/ignore/io/ErrorHandlingLoader.java
roesslerj: This is consistent with `FilterPreserveLineLoader`. Either change both or keep it that way... ;-)

diba1013 (approved):

PR #273, opened 2019-06-08
Remove retest ID provider util method

beatngu13:
Isn't being used outside tests anyway.


roesslerj (commented):
I'd happily approve this PR. But unfortunately, the real work is to replace the usage of those methods in downstream projects... until I have seen that, I'd rather not approve...

beatngu13:
I searched all downstream projects (including recheck-web, review, etc.), this particular method seems to be used nowhere. It's dead code. Removing the util is another thing I would say.

roesslerj:
> I searched all downstream projects (including recheck-web, review, etc.), this particular method seems to be used nowhere. It's dead code. Removing the util is another thing I would say.

Ok. On your responsibility...

roesslerj (approved):

PR #277, opened 2019-06-13
Adapt test case finder API

### LINKED ISSUE #205: Parameterized tests are overwritten
Suppose the following recheck-web test:

```java
@ParameterizedTest
@MethodSource( "com.github.beatngu13.visualtesting.util.DriverFactory#getAll" )
void test(WebDriver driver) throws Exception {
	re.startTest();
	driver.get("https://example.com/");
	re.check(driver, "example");
	re.capTest();
}
```

If the user doesn't provide a test name via `startTest(String)` and uses the default `startTest()`, we simply overwrite the previous GM as all parameterized test cases have the same method name.

We should try to recognize this and then at least append a suffix.

beatngu13:
Two reviewers on purpose, please don't approve prematurely.

Resolves #205.

### DIFF
--- a/src/main/java/de/retest/recheck/TestCaseFinder.java
+++ b/src/main/java/de/retest/recheck/TestCaseFinder.java
@@ -81,12 +81,9 @@ public class TestCaseFinder {
 	}
 
 	public static StackTraceElement findTestCaseMethodInStack( final StackTraceElement[] trace ) {
-		boolean inTestCase = false;
-		for ( int i = 0; i < trace.length; i++ ) {
-			if ( isTestCase( trace[i] ) ) {
-				inTestCase = true;
-			} else if ( inTestCase ) {
-				return trace[i - 1];
+		for ( final StackTraceElement element : trace ) {
+			if ( isTestCase( element ) ) {
+				return element;
 			}
 		}
 		return null;
--- a/src/main/java/de/retest/recheck/TestCaseFinder.java
+++ b/src/main/java/de/retest/recheck/TestCaseFinder.java
@@ -15,6 +15,11 @@ public class TestCaseFinder {
 
 	private static final Logger logger = LoggerFactory.getLogger( TestCaseFinder.class );
 
+	/**
+	 * Delimiter to separate methods names from their invocation count.
+	 */
+	public static final String DELIMITER = "_";
+
 	private TestCaseFinder() {}
 
 	/*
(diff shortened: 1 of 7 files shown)


varlog23 (commented):
Everything works as expected. Could not find any flaws with this approach.

src/main/java/de/retest/recheck/TestCaseFinder.java
```diff
@@ -32,51 +37,128 @@ private TestCaseFinder() {}
...
+	 * @return Test case information for the test case method in the given stack trace.
+	 */
+	public static TestCaseInformation findTestCaseMethodInStack( final StackTraceElement[] trace ) {
+		for ( final StackTraceElement element : trace ) {
+			final TestCaseAnnotationType type = determineTestCaseAnnotationType( element );
+			if ( type == TestCaseAnnotationType.NORMAL ) {
+				return new TestCaseInformation( element, type, 1 );
+			}
+			if ( type == TestCaseAnnotationType.REPEATABLE ) {
+				final int count = repeatableTestCaseAnnotationsCount.merge( element, 1, Math::addExact );
```
beatngu13: Use `Integer#sum( int, int )`.

src/main/java/de/retest/recheck/TestCaseFinder.java
```diff
@@ -32,51 +37,128 @@ private TestCaseFinder() {}
...
 		}
 
 		final Annotation[] annotations = method.getAnnotations();
 		for ( final Annotation annotation : annotations ) {
 			final String annotationName = annotation.annotationType().getName();
 			if ( testCaseAnnotations.contains( annotationName ) ) {
-				return true;
+				return TestCaseAnnotationType.NORMAL;
+			}
+			if ( repeatableTestCaseAnnotations.contains( annotationName ) ) {
```
beatngu13, line 165: Change order.

src/main/java/de/retest/recheck/TestCaseFinder.java
```diff
@@ -32,51 +37,128 @@ private TestCaseFinder() {}
...
+	private static final Set<String> repeatableTestCaseAnnotations = new HashSet<>( Arrays.asList( //
+			// JUnit Vintage (v4)
+			"org.junit.experimental.theories.Theory", //
+			// JUnit Jupiter (v5)
+			"org.junit.jupiter.api.RepeatedTest", //
+			"org.junit.jupiter.params.ParameterizedTest" ) );
+
+	private static final Map<StackTraceElement, Integer> repeatableTestCaseAnnotationsCount = new HashMap<>();
+
+	private TestCaseFinder() {}
```
beatngu13: To singleton.

src/main/java/de/retest/recheck/TestCaseFinder.java
beatngu13: I think I'll keep it as it is, natural order. Having two annotations is just wrong usage I guess.

diba1013 (approved):

PR #291, opened 2019-06-25
Enable cloud persistence for rehub

modulo11:


src/main/java/de/retest/recheck/Rehub.java
```diff
@@ -18,6 +18,7 @@ public static void init() {
 		if ( !auth.isAuthenticated( getToken() ) ) {
 			auth.login( new RehubAuthenticationHandler() );
 		}
+		System.setProperty( Properties.REHUB_ENABLED, "true" );
```
beatngu13: How about a test for this? `init()` should set rehub system property. This can either be an integration test via `RecheckImpl` or a unit test for `Rehub`.

beatngu13 (approved):

PR #295, opened 2019-06-26
Lo and behold, path is also an invisible attribute...

roesslerj:


src/main/resources/filter/web/invisible-attributes.filter
```diff
@@ -1,6 +1,7 @@
 # Filter file for recheck with invisible and mostly irrelevant attributes 
 
 # Class can be ignored, because we track individual CSS attributes
```
beatngu13: Doesn't this comment belong to `attribute=class`? Is it still needed at all?

beatngu13 (changes requested):

src/main/resources/filter/web/invisible-attributes.filter
roesslerj: Wasn't that change, but Good Catch ;-)

beatngu13 (approved):

src/main/resources/filter/web/invisible-attributes.filter
beatngu13: Yeah, I overlooked this earlier… 🙈

PR #305, opened 2019-07-03
Adapt the element constructor

ivonx:


src/main/java/de/retest/recheck/ui/descriptors/Element.java
```diff
@@ -68,7 +68,7 @@ protected Element() {
 			final Attributes attributes, final Screenshot screenshot ) {
 		RetestIdUtil.validate( retestId, identifyingAttributes );
 		this.retestId = retestId;
-		this.parent = Objects.requireNonNull( parent, "Parent must not be null" );
```
beatngu13, line 71: Why should we do this? The only element that accepts a null parent is `RootElement`…

beatngu13: @ivonx just told me that this is for updating recheck within retest. However, I believe that we should adapt retest for recheck, not vice versa. This is "good" code and somewhat an invariant we need in recheck and downstream projects. Therefore, I would suggest to decline this. @roesslerj you said it's OK?

roesslerj: Ah, sorry ... that was a misunderstanding. I thought ivonx was talking about test code within retest. I am also in favor of declining.

roesslerj: And on top of that, it will not help us with migration if we change this in recheck v1.3.0...

roesslerj:
Sorry ... this was a misunderstanding. Please fix this in the test code.

PR #316, opened 2019-07-10
Feature/improve auth

modulo11:


beatngu13 (approved):
I'm good, but Sonar complains.

PR #324, opened 2019-07-16
This adds Git branch and commit to the header

roesslerj:
How do we test that? It works as is show by executing the `GitUtil.main` method...


modulo11:
Why is that needed? What if people do not use git?
This idea should be presented to all developers.

beatngu13:
I assume the intend is that if the same test is executed on multiple branches, it is supposed to help the user to differentiate the reports?

I have the same concerns as Joe, what if the user not uses Git, but SVN, Mercurial, … or no VCS at all? All that Applitools is doing is a simple timestamp with metadata for OS, browser, viewport etc. 

grafik

Not as beauty as branch names, but much easier to implement, requires no VCS detection, no shell command execution, …

modulo11:
As a stated several times, we need to discuss which meta data we want in general.

PR #336, opened 2019-07-31
Feature/fix comparison

alig01:
Changes in the absolute-outline were previously not recognized by our equals method


roesslerj:
This looks good. But just to make sure: Ignoring `absoluteOutline` via property still works, right?

alig01:
Yes it should still work.

roesslerj:
As far as I can see, that should still work... but can you make sure (i.e. have a short - if manual - test for that)?

roesslerj (approved):

PR #338, opened 2019-07-31
We can even do the opposite, and filter content but no CSS...

roesslerj:


src/main/resources/filter/web/content.filter
```diff
@@ -0,0 +1,3 @@
+# Filter file for recheck that will filter content
+
+attribute=text
```
diba1013, line 3: I don't quite get the commit message ... What are you trying to do?
As always ... We cannot just simply add `.filter`-files. They need to be registered.

diba1013 (changes requested):

alig01 (approved):

diba1013 (approved):

PR #347, opened 2019-08-02
Refactor

beatngu13:


src/main/java/de/retest/recheck/RecheckImpl.java
```diff
@@ -215,8 +217,7 @@ String getDifferencesErrorMessage( final TestReplayResult finishedTestResult,
 				+ "You can review the details by using our CLI (https://github.com/retest/recheck.cli/) or GUI (https://retest.de/review/).\n" //
 				+ "\n" //
 				+ numChecks + " check(s) in '" + suiteName + "' found the following difference(s):\n" //
-				+ allDiffs //
-				+ insertedDeletedDiffs.toString();
+				+ allDiffs + insertedDeletedDiffs.toString();
```
diba1013: I think there should be a "\n", since the printer does not ensure a newline at the end.

src/test/java/de/retest/recheck/RecheckImplTest.java
```diff
@@ -102,6 +108,58 @@ void calling_with_no_GM_should_produce_better_error_msg( @TempDir final Path roo
...
+
+		final Element absent = null;
+		final Element present = mock( Element.class );
+		when( present.getIdentifyingAttributes() ).thenReturn( identifyingAttributes );
+
+		final InsertedDeletedElementDifference insertion =
+				InsertedDeletedElementDifference.differenceFor( absent, present );
+		final InsertedDeletedElementDifference deletion =
+				InsertedDeletedElementDifference.differenceFor( present, absent );
+		final LeafDifference ignore = mock( LeafDifference.class );
```
diba1013: Could you add a comment on why you added this `ignore`? Or maybe even make a separate test?

diba1013 (changes requested):

src/main/java/de/retest/recheck/RecheckImpl.java
beatngu13: Makes sense, but let me first write a test for it.

src/main/java/de/retest/recheck/AssertionMessage.java
```diff
@@ -0,0 +1,61 @@
...
+
+import de.retest.recheck.printer.TestReplayResultPrinter;
+import de.retest.recheck.report.ActionReplayResult;
+import de.retest.recheck.report.TestReplayResult;
+import de.retest.recheck.ui.diff.InsertedDeletedElementDifference;
+import de.retest.recheck.ui.diff.LeafDifference;
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+class AssertionMessage {
```
diba1013: `AssertionMessage` is really generic and in my opinion does two things. Do we want to have specific error message classes for both `getNoGoldenMasterErrorMessage` and `getDifferencesErrorMessage`?

diba1013 (changes requested):

src/main/java/de/retest/recheck/AssertionMessage.java
beatngu13: We can use `RecheckAssertionMessage` or `RecheckCapMessage` if you want to?

However, I wouldn't go for separate classes in this case. It already moves complexity out of `RecheckImpl` into a class, which is solely concerned with creating a proper assertion message for the given test result data. Therefore, testing became way easier.

I don't think there is much benefit in further splitting this up into separate types. Either `RecheckImpl` would have to decide which one to create, or you have something like a factory that does this.

diba1013: > We can use `RecheckAssertionMessage` or `RecheckCapMessage` if you want to?

Yes, sure, just to be more precise. Although I would prefer `RecheckCapMessage` to make sure it is not misused in some other place.

> I don't think there is much benefit in further splitting this up into separate types

My general rule of thumb (with few exceptions and respective [builder](https://dzone.com/articles/too-many-parameters-java-1)) is not to have a class/method with more than 3 parameters as indicated by https://rules.sonarsource.com/java/RSPEC-107. Sadly this is not the only class in our code.
 _So maybe a step at a time._

beatngu13: I've adapted the class name, definitely better now.

I also agree that the 5 ctor args that are currently used aren't beautiful, but how would splitting up the implementation help here? `getDifferencesErrorMessage()` needs all of them.

diba1013: Well, maybe address this separately.

diba1013 (approved):

PR #350, opened 2019-08-06
Feature/ignelem should ign all attr diffs

roesslerj:
I have no idea why this was missing ... but it is specified in the docs: https://docs.retest.de/recheck/how-ignore-works/

### DIFF
--- a/src/main/java/de/retest/recheck/review/ignore/ElementFilter.java
+++ b/src/main/java/de/retest/recheck/review/ignore/ElementFilter.java
@@ -26,7 +26,7 @@ public class ElementFilter implements Filter {
 
 	@Override
 	public boolean matches( final Element element, final AttributeDifference attributeDifference ) {
-		return false;
+		return matcher.test( element );
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


src/main/java/de/retest/recheck/review/ignore/ElementFilter.java
```diff
@@ -26,7 +26,7 @@ public boolean matches( final Element element ) {
 
 	@Override
 	public boolean matches( final Element element, final AttributeDifference attributeDifference ) {
-		return false;
+		return matcher.test( element );
```
diba1013, line 29: That was made intentionally and that is why we added both methods `matches(Element)` and `matches(Element, AttributeDifference)`. The task of a filter is to describe the matching based on the available data. An `ElementFilter` is only capable of ignoring elements, not attribute difference, since it has not that much information present. Analog the `ElementAttributeFilter` is not capable of filtering `matches(Element)` since it needs more information than there information is.

Thus the filter is not responsible for attribute filters to check for the element. The element in this method should only be used for context and not for main purpose (thats what `match(Element` is for). The calling implementation, however, need to verify that both methods are called in **every** circumstance.

Let's discuss this tomorrow.

diba1013 (changes requested):

diba1013:
See comment, as this is intentional behavior. This should be fixed at the calling side.

PR #355, opened 2019-08-09
Yet another invisible attribute...

roesslerj:


src/main/resources/filter/web/invisible-attributes.filter
```diff
@@ -12,6 +12,7 @@ attribute=jsdata
 attribute=jsaction
 attribute=cd_frame_id_
 attribute=transform
+attribute=tabindex
```
diba1013, line 15: tabindex as specified by [w3c ](https://html.spec.whatwg.org/multipage/interaction.html#attr-tabindex) or tab-index as in [getAllElementsByPath.js](https://github.com/retest/recheck-web/blob/master/src/main/resources/javascript/getAllElementsByPath.js#L74)?

I suggest first, since it is the standard. We need to unify these as in writing a migrator and such (separate task, I know).

roesslerj: Right. So the first is currently in this PR. Anything else you want? :-)

diba1013 (approved):

PR #364, opened 2019-08-19
Feature/check recheck version

roesslerj:


src/main/java/de/retest/recheck/ioerror/ReTestIOException.java
```diff
@@ -2,24 +2,27 @@
...
-import de.retest.recheck.util.FileUtil;
+import java.net.URI;
 
 public class ReTestIOException extends IOException {
 
 	private static final long serialVersionUID = 1L;
-	private final File file;
+	private final URI location;
 
 	public ReTestIOException( final File file, final Throwable throwable ) {
```
beatngu13: Since this is an internal exception and we now use `URI` instead of `File` (see e.g. `FileUtil`), I would remove this ctor to reduce the amount of code we have to maintain. Same goes for `ReTestLoadException` and `ReTestSaveException`.

src/main/java/de/retest/recheck/persistence/IncompatibleReportVersionException.java
```diff
@@ -0,0 +1,29 @@
...
+public class IncompatibleReportVersionException extends ReTestLoadException {
+
+	private static final long serialVersionUID = 1L;
+
+	private final String writerVersion;
+	private final String readerVersion;
+
+	public IncompatibleReportVersionException( final String writerVersion, final String readerVersion,
+			final URI location, final Throwable throwable ) {
+		super( location, "Incompatible retest versions: report was written with " + writerVersion + ", but read with "
```
beatngu13: Shouldn't this be:

```suggestion
		super( location, "Incompatible recheck versions: report was written with " + writerVersion + ", but read with "
```

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -9,19 +9,34 @@
...
+	private final Kryo kryo;
+	private final String version;
+
+	public KryoPersistence() {
+		kryo = createKryo();
+		version = VersionProvider.RETEST_VERSION;
+	}
+
+	// ONLY FOR TESTING
+	public KryoPersistence( final Kryo kryo, final String version ) {
```
beatngu13: Can be package private.

src/test/java/de/retest/recheck/persistence/bin/KryoPersistenceTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+class KryoPersistenceTest {
+
+	@Test
+	void roundtrip_should_work( @TempDir final Path temp ) throws IOException {
+		final Path file = temp.resolve( "test.test" );
+		Files.createFile( file );
+		final URI identifier = file.toUri();
+
+		final KryoPersistence<de.retest.recheck.test.Test> kryoPersistence = new KryoPersistence<>();
```
beatngu13, line 32: Why not use a "real" `TestReport` instance like in the `KryoPersistenceTest` in review?

src/test/java/de/retest/recheck/persistence/bin/KryoPersistenceTest.java
```diff
@@ -0,0 +1,60 @@
...
+		final ArrayList<String> tests = new ArrayList<>();
```
beatngu13, line 33: You can remove line 33 and do:

```suggestion
		final List<String> tests = Arrays.asList( "../test.test" );
```

Same in the other test.

beatngu13 (changes requested):

src/test/java/de/retest/recheck/persistence/bin/KryoPersistenceTest.java
roesslerj: Nope, because this returns an ArrayList that is not persistable... 👎 
I know, because this is exactly the code I had before.

src/main/java/de/retest/recheck/ioerror/ReTestIOException.java
roesslerj: OK

src/main/java/de/retest/recheck/persistence/IncompatibleReportVersionException.java
roesslerj: Ok

src/test/java/de/retest/recheck/persistence/bin/KryoPersistenceTest.java
roesslerj: Because it uses a lot of `builder` methods, that are not available in recheck (yet).

beatngu13 (approved):

PR #372, opened 2019-08-22
Fix pixel diff save

beatngu13:
Both float (`pixel-diff=5.0`) and integer (`pixel-diff=5`) is allowed for specifying allowed pixel diffs in a `recheck.ignore`. However, as we internally always treat it as a double, we have to remember how it has been specified such that we don't manipulate existing ignore files.

### DIFF
--- a/src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
+++ b/src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
@@ -9,8 +9,10 @@ import de.retest.recheck.ignore.Filter;
 import de.retest.recheck.review.ignore.io.RegexLoader;
 import de.retest.recheck.ui.descriptors.Element;
 import de.retest.recheck.ui.diff.AttributeDifference;
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
+@Getter
 @Slf4j
 public class PixelDiffFilter implements Filter {
 
--- a/src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
+++ b/src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
@@ -19,14 +19,14 @@ public class PixelDiffFilter implements Filter {
 	private static final String PIXEL = "px";
 
 	/**
-	 * Indicates whether {@link #pixelDiff} is specified as float ({@code true}) or integer ({@code false}). Although
+	 * Indicates whether {@link #pixelDiff} is specified as double ({@code true}) or integer ({@code false}). Although
 	 * internally it is always treated as a double, this is important for serialization.
 	 */
-	private final boolean specifiedAsFloat;
+	private final boolean specifiedAsDouble;
 	private final double pixelDiff;
 
-	public PixelDiffFilter( final boolean specifiedAsFloat, final double pixelDiff ) {
-		this.specifiedAsFloat = specifiedAsFloat;
+	public PixelDiffFilter( final boolean specifiedAsDouble, final double pixelDiff ) {
+		this.specifiedAsDouble = specifiedAsDouble;
 		this.pixelDiff = pixelDiff;
 	}
 
@@ -84,7 +84,7 @@ public class PixelDiffFilter implements Filter {
 
 	@Override
 	public String toString() {
-		final String value = specifiedAsFloat ? Double.toString( pixelDiff ) : Integer.toString( (int) pixelDiff );
+		final String value = specifiedAsDouble ? Double.toString( pixelDiff ) : Integer.toString( (int) pixelDiff );
 		return String.format( PixelDiffFilterLoader.FORMAT, value );
 	}
 
@@ -101,9 +101,9 @@ public class PixelDiffFilter implements Filter {
 		@Override
 		protected PixelDiffFilter load( final MatchResult regex ) {
 			final String value = regex.group( 1 );
-			final boolean specifiedAsFloat = value.contains( "." );
+			final boolean specifiedAsDouble = value.contains( "." );
 			final double pixelDiff = Double.parseDouble( value );
-			return new PixelDiffFilter( specifiedAsFloat, pixelDiff );
+			return new PixelDiffFilter( specifiedAsDouble, pixelDiff );
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
```diff
@@ -76,7 +82,8 @@ private static String clean( final String str ) {
 
 	@Override
 	public String toString() {
-		return String.format( PixelDiffFilterLoader.FORMAT, pixelDiff );
+		final String value = specifiedAsFloat ? Double.toString( pixelDiff ) : Integer.toString( (int) pixelDiff );
+		return String.format( PixelDiffFilterLoader.FORMAT, value );
```
diba1013, line 88: I'm quite unsure, if I want specific formatting for serialization here or in the `Loader#save` (which per default delegates to `toString()`). 
Pro:
+ It separates the logic from the representation, since the filter actually does not care for the format.
+ The loader is responsible for the formatting and thus ensures the proper formatting upon saving.

Contra:
- It introduces somewhat duplicate code.
- We still need to have some kind of indicator inside/with the filter class (let it be a flag argument or instance-of check) for the loader to distinguish. See my comment above.

Ideas?

src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
```diff
@@ -16,9 +16,15 @@
...
 
+	/**
+	 * Indicates whether {@link #pixelDiff} is specified as float ({@code true}) or integer ({@code false}). Although
+	 * internally it is always treated as a double, this is important for serialization.
+	 */
+	private final boolean specifiedAsFloat;
 	private final double pixelDiff;
 
-	public PixelDiffFilter( final double pixelDiff ) {
+	public PixelDiffFilter( final boolean specifiedAsFloat, final double pixelDiff ) {
```
diba1013: I'm not very fond of these flag arguments (see Clean Code). A solution would either to 
- pass in a specific Formatter (e.g. `DoubleFunction<String> formatter`) - _also not so good_ 
- have a inner class `IntPixelDiffFilter` and `DoublePixelDiffFilter` which uses `PixelDiffFilter` as parent while overwriting `toString` or similar - _introduces verbosity_

What do you think?

src/test/java/de/retest/recheck/review/ignore/PixelDiffFilterTest.java
```diff
@@ -14,7 +14,7 @@
 class PixelDiffFilterTest {
 
 	double pixelDiff = 5.0;
-	Filter cut = new PixelDiffFilter( pixelDiff );
+	Filter cut = new PixelDiffFilter( false, pixelDiff );
```
diba1013, line 17: Just for safety, do we want a `new PixelDiffFilter( true, pixelDiff )` for specific checks? Or do you think, we should expand the `PixelDiffFilterLoaderTest` to have it call `save` to ensure formatting?

Also, when do you think, we should use a `@BeforeEach`-method?

beatngu13: > Just for safety, do we want a `new PixelDiffFilter( true, pixelDiff )` for specific checks? Or do you think, we should expand the `PixelDiffFilterLoaderTest` to have it call `save` to ensure formatting?

It is currently tested via the `LoadAndSaveFilterWorkerTest`, but I can introduce a separate unit test for the `toString()` method if you want?

> Also, when do you think, we should use a `@BeforeEach`-method?

It boils down to:

```java
class MyTest {
	Object o = new Object();
}
```

vs.:

```java
class MyTest {
	Object o;

	@BeforeEach
	void setUp() {
		o = new Object();
	}
}
```

What is the benefit of the latter if there is no complex setup (e.g. with mocking)? Although one can change the [lifecycle of a JUnit 5 test class](https://blog.codefx.org/libraries/junit-5-basics/#Test-Class-Lifecycle), this probably never happens globally but rather on demand per class.

Our [wiki article](https://retest.atlassian.net/wiki/spaces/RET/pages/160464907/Testing) also suggests no general rule here at the moment, so maybe we should gather some other opinions as well as best practices and discuss this in the retro?

src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
beatngu13: I see your point, but the "flag argument rule" targets methods/functions, not constructors. I think both are different as constructing/configuring an object can be more complex than calling one of its methods.

In fact, all filters currently violate the SoC principle to a certain degree as they not just know how to filter, but also how they are going to be serialized via the `toString()` method. IMHO having separate (floating and integer) classes or a special formatter isn't worth the effort, but I'm open.

src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
beatngu13: Valid arguments, but I think this PR is the wrong place to discuss this. It fixes a bug, and as you already said, it currently delegates to `toString()` so it follows the given abstractions. Changing these abstractions should be done separately I would say?

diba1013: Yes, sure.

src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
diba1013: > In fact, all filters currently violate the SoC principle to a certain degree

It currently just sticks out (more), since it especially needs special information regarding formatting, although its (logic) implementation does not care about integers.
We can discuss this at a later point, once we do implement SoC.

src/test/java/de/retest/recheck/review/ignore/PixelDiffFilterTest.java
diba1013: > It is currently tested via the `LoadAndSaveFilterWorkerTest`, but I can introduce a separate unit test for the `toString()` method if you want?

Sure, it is up to you, if you want to test the `PixelDiffFilter#toString` or the respective `PixelDiffFilterLoaderTest`. Mainly that we have a more direct test with an appropriate description for this fix.

IMHO, `LoadShouldIgnoreWorkerTest` should be an integration test (but that's separate, too).

diba1013 (changes requested):

src/test/java/de/retest/recheck/review/ignore/PixelDiffFilterTest.java
beatngu13: OK, will add a unit test, we can still keep the integration test. 👍

diba1013 (approved):

PR #375, opened 2019-08-25
Feature/options filter

roesslerj:
As far as I understood, this should be a first step towards solving https://github.com/retest/recheck-web/issues/326:
"As long as there's a way to ignore on demand, via code I'm fine."

Anyways this is a good idea. 

Now it's great that we filter only for assertions, but still report any differences in the report ... making this a sensible and unproblematic change.

### DIFF
--- a/src/main/java/de/retest/recheck/RecheckOptions.java
+++ b/src/main/java/de/retest/recheck/RecheckOptions.java
@@ -1,24 +1,91 @@
 package de.retest.recheck;
 
-import lombok.Builder;
-import lombok.Getter;
-import lombok.experimental.SuperBuilder;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
-@SuperBuilder
+import de.retest.recheck.ignore.CompoundFilter;
+import de.retest.recheck.ignore.Filter;
+import de.retest.recheck.ignore.RecheckIgnoreUtil;
+import lombok.Getter;
+
 @Getter
 public class RecheckOptions {
 
-	@Builder.Default
-	private final FileNamerStrategy fileNamerStrategy = new MavenConformFileNamerStrategy();
+	private final FileNamerStrategy fileNamerStrategy;
+	private final String suiteName;
+	private final boolean reportUploadEnabled;
+	private final Filter filter;
 
-	@Builder.Default
-	private final String suiteName = null;
-
-	@Builder.Default
-	private final boolean reportUploadEnabled = false;
+	public RecheckOptions( final FileNamerStrategy fileNamerStrategy, final String suiteName,
+			final boolean reportUploadEnabled, final Filter filter ) {
+		this.fileNamerStrategy = fileNamerStrategy;
+		this.suiteName = suiteName;
+		this.reportUploadEnabled = reportUploadEnabled;
+		this.filter = filter;
+	}
 
 	public String getSuiteName() {
 		return suiteName == null ? fileNamerStrategy.getTestClassName() : suiteName;
 	}
 
+	public static RecheckOptionsBuilder builder() {
+		return new RecheckOptionsBuilder();
+	}
+
+	public static class RecheckOptionsBuilder {
+		private FileNamerStrategy fileNamerStrategy = new MavenConformFileNamerStrategy();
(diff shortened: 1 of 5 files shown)


src/main/java/de/retest/recheck/RecheckOptions.java
```diff
@@ -1,10 +1,15 @@
...
+import de.retest.recheck.ignore.CompoundFilter;
+import de.retest.recheck.ignore.Filter;
+import de.retest.recheck.ignore.RecheckIgnoreUtil;
+import de.retest.recheck.review.GlobalIgnoreApplier;
 import lombok.Builder;
 import lombok.Getter;
-import lombok.experimental.SuperBuilder;
 
-@SuperBuilder
+@Builder
```
beatngu13: IIRC this breaks compatability with `RecheckWebOptions`.

src/main/java/de/retest/recheck/RecheckOptions.java
```diff
@@ -15,10 +20,48 @@
...
+	private final boolean reportUploadEnabled = false;
+
+	@Builder.Default
+	private final Filter filter = new CompoundFilter( RecheckIgnoreUtil.loadRecheckIgnore() );
 
 	public String getSuiteName() {
 		return suiteName == null ? fileNamerStrategy.getTestClassName() : suiteName;
 	}
 
+	public void addFilter( final Filter filter ) {
```
beatngu13: IMHO our options should be immutable after construction and only offer getters. Therefore, I would suggest a builder method for a list (or varargs) of `Filter`s.

src/main/java/de/retest/recheck/ignore/CompoundFilter.java
```diff
@@ -37,4 +47,8 @@ public boolean matches( final Element element, final AttributeDifference attribu
 		return filters;
 	}
 
+	public void add( final Filter filter ) {
```
beatngu13: I think we should strive for immutability here too. If `RecheckOptions` allows something like:

```java
RecheckOptions opts = RecheckOptions.builder()
        .filter( f0, f1, f2)
        .build();
```

I think there is no need for this.

src/main/java/de/retest/recheck/ignore/InverseFilter.java
```diff
@@ -0,0 +1,23 @@
+package de.retest.recheck.ignore;
+
+import de.retest.recheck.ui.descriptors.Element;
+import de.retest.recheck.ui.diff.AttributeDifference;
+
+public class InverseFilter implements Filter {
```
beatngu13: Puh, that's a though decision. While Git allows this, there are three things two consider:

1. This breaks our rule "filters are only additive", now they can be conflicts I guess.
2. Isn't blacklist testing something we discourage?
3. There should be a way to also do this from `recheck.ignore`.

beatngu13 (changes requested):

src/main/java/de/retest/recheck/RecheckOptions.java
roesslerj: As discussed...

src/main/java/de/retest/recheck/ignore/InverseFilter.java
roesslerj: Removed

src/main/java/de/retest/recheck/RecheckOptions.java
```diff
@@ -1,24 +1,91 @@
...
-	private boolean reportUploadEnabled = false;
+	public RecheckOptions( final FileNamerStrategy fileNamerStrategy, final String suiteName,
+			final boolean reportUploadEnabled, final Filter filter ) {
+		this.fileNamerStrategy = fileNamerStrategy;
+		this.suiteName = suiteName;
+		this.reportUploadEnabled = reportUploadEnabled;
+		this.filter = filter;
+	}
 
 	public String getSuiteName() {
```
beatngu13: We don't have to trick Lombok anymore, so we should be able to remove this getter and initialize the default within the builder as follows:

```java
private String suiteName = fileNamerStrategy.getTestClassName();
```

beatngu13 (approved):

PR #379, opened 2019-08-27
Add test and treat old versions and matching versions correctly

roesslerj:


src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -74,14 +76,22 @@ public void save( final URI identifier, final T element ) throws IOException {
 		}
 	}
 
+	private final String VERSION_CHARS = "[\\w\\.\\{\\}\\$]+";
+
 	@SuppressWarnings( "unchecked" )
 	@Override
 	public T load( final URI identifier ) throws IOException {
-		String writerVersion = "unknown";
+		String writerVersion = OLD_RECHECK_VERSION;
```
beatngu13: Since you have `null` check in line 92 anyway: wouldn't it be more intuitive to have here:

```suggestion
		String writerVersion = null;
```

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -74,14 +76,22 @@ public void save( final URI identifier, final T element ) throws IOException {
...
 		try ( final Input input = new Input( Files.newInputStream( Paths.get( identifier ) ) ) ) {
 			writerVersion = input.readString();
 			return (T) kryo.readClassAndObject( input );
 		} catch ( final KryoException e ) {
+			if ( version.equals( writerVersion ) ) {
```
beatngu13, line 89: Optional and more of a matter of taste: maybe it is helpful to wrap those boolean expressions inside methods? For example:

```java
if ( isSameVersion( writerVersion ) ) {
    throw e;
}
if ( isUnknownFormat( writerVersion ) ) {
    writerVersion = OLD_RECHECK_VERSION;
}

File: src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java [hunk 3]
```diff
@@ -74,14 +76,22 @@ public void save( final URI identifier, final T element ) throws IOException {
 		}
 	}
 
+	private final String VERSION_CHARS = "[\\w\\.\\{\\}\\$]+";
```
beatngu13: Since it is a constant and a regex:

```suggestion
	private static final String VERSION_CHARS_REGEX = "[\\w\\.\\{\\}\\$]+";
```

You could also compile the regex once (via `Pattern.compile( String )`) and then use it for matching.

beatngu13 (changes requested):

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
roesslerj: I think wrapping `version.equals(` is less readable, as it hides otherwise easily understandable code.

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
roesslerj: But then the other part is less readable, because of null-check...

beatngu13 (approved):

beatngu13:
Oops, this change seems to cause test failures.

beatngu13 (approved):

beatngu13:
I think you have to rebase manually since @rebazer appears to be a bit lazy these days…

PR #380, opened 2019-08-27
Feature/add utility for element difference

diba1013:
This is needed inside the printers to remove verbosity on printing. Currently this is only implemented half ways inside `ActionReplayResultPrinter`, but has no use until `ElementDifferencePrinter` can handle all possible differences from `ActionReplayResult`.

### DIFF
--- a/src/test/java/de/retest/recheck/ui/diff/ElementDifferenceTest.java
+++ b/src/test/java/de/retest/recheck/ui/diff/ElementDifferenceTest.java
@@ -1,0 +1,43 @@
+package de.retest.recheck.ui.diff;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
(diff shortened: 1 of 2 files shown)


src/test/java/de/retest/recheck/ui/diff/ElementDifferenceTest.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+import de.retest.recheck.ui.descriptors.Element;
+
+class ElementDifferenceTest {
+
+	@Test
+	void hasAttributesDifferences() {
```
beatngu13: Should pattern or at least `testHasAnyDifference()`.

beatngu13 (changes requested):

beatngu13 (approved):

PR #381, opened 2019-08-27
Feature/fix report filter if no diff

diba1013:
When using the filter mechanism with no differences (i.e. `StateDifference` is null), this method throws an `NPE`.

### DIFF
--- a/src/test/java/de/retest/recheck/report/ActionReplayResultTest.java
+++ b/src/test/java/de/retest/recheck/report/ActionReplayResultTest.java
@@ -12,6 +12,7 @@ import org.mockito.Mockito;
 
 import de.retest.recheck.execution.RecheckDifferenceFinder;
 import de.retest.recheck.ignore.Filter;
+import de.retest.recheck.report.action.ActionReplayData;
 import de.retest.recheck.review.ignore.AttributeFilter;
 import de.retest.recheck.ui.DefaultValueFinder;
 import de.retest.recheck.ui.Path;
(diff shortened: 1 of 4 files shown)


src/test/java/de/retest/recheck/report/ActionReplayResultTest.java
```diff
@@ -109,4 +111,25 @@ void getDiffrences_should_ignore_attributes_changes() {
...
+				ActionReplayResult.createActionReplayResult( data, null, null, difference, 0L, sutState );
+
+		assertThat( result.hasDifferences() ).isFalse();
+	}
+
+	@Test
+	void factory_method_should_return_no_difference_if_state_is_empty() {
+		final ActionReplayData data = mock( ActionReplayData.class );
+		final SutState sutState = mock( SutState.class );
+		final StateDifference difference = mock( StateDifference.class ); // This difference should be empty vs null
```
beatngu13, line 129: Is this a "valid" situation? Having an empty state difference? Shouldn't it be `null` or differences, nothing in between?

diba1013: Well, it was checked there before, so I added this test.

But ideally, we would get rid of the `null`s and use empty `StateDifference` to remove the tedious `null`-checks.

beatngu13: Allright, let me approve this one.

beatngu13 (approved):

PR #382, opened 2019-08-27
Feature/adapt no golden master

diba1013:
Was there any reason on why this fails if only all results are `NoGoldenMasterActionReplayResult`?


roesslerj (dismissed):

roesslerj (changes requested):
This should be (as the method name suggests) `allMatch`.

diba1013:
Seems like Github is a bit confused, as I cannot request a new review.

roesslerj (approved):

PR #383, opened 2019-08-27
Feature/keep golden master on filter

diba1013:
We destroyed the `NoGoldenMasterActionReplayResult` when filtering

Should be able to merge after #382

### DIFF
--- a/src/main/java/de/retest/recheck/report/TestReportFilter.java
+++ b/src/main/java/de/retest/recheck/report/TestReportFilter.java
@@ -5,6 +5,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import de.retest.recheck.NoGoldenMasterActionReplayResult;
 import de.retest.recheck.ignore.Filter;
 import de.retest.recheck.report.action.ActionReplayData;
 import de.retest.recheck.ui.actions.ExceptionWrapper;
@@ -51,6 +52,10 @@ public class TestReportFilter {
 	}
 
 	static ActionReplayResult filter( final ActionReplayResult actionReplayResult, final Filter filter ) {
+		if ( actionReplayResult instanceof NoGoldenMasterActionReplayResult ) {
+			return actionReplayResult;
+		}
+
 		final ActionReplayData data = ActionReplayData.withTarget( actionReplayResult.getDescription(),
 				actionReplayResult.getTargetComponent(), actionReplayResult.getGoldenMasterPath() );
 		final ExceptionWrapper error = actionReplayResult.getThrowableWrapper();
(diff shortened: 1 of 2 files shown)


src/test/java/de/retest/recheck/report/TestReportFilterTest.java
```diff
@@ -177,6 +178,22 @@ void test_replay_result_should_be_filtered_properly() throws Exception {
 		assertThat( differences ).containsExactly( notFilterMe );
 	}
 
+	@Test
+	void test_replay_result_should_keep_golden_master_exceptions() {
+		final NoGoldenMasterActionReplayResult r1 = mock( NoGoldenMasterActionReplayResult.class );
```
beatngu13: Find it a bit confusing that replay results are named here `r`-something, whereas below `result` and `filter`. How about:

* `r1` => `noGoldenMasterActionResult`
* `r2` => `stateDiffActionResult`
* `result` => `testResult`
* `filter` => `filteredTestResult`

src/main/java/de/retest/recheck/report/TestReportFilter.java
```diff
@@ -45,7 +46,11 @@ static TestReplayResult filter( final TestReplayResult testReplayResult, final F
 		final TestReplayResult newTestReplayResult =
 				new TestReplayResult( testReplayResult.getName(), testReplayResult.getTestNr() );
 		for ( final ActionReplayResult actionReplayResult : testReplayResult.getActionReplayResults() ) {
-			newTestReplayResult.addAction( filter( actionReplayResult, filter ) );
+			if ( actionReplayResult instanceof NoGoldenMasterActionReplayResult ) {
```
beatngu13: Isn't the actual problem that `filter( ActionReplayResult, Filter )` filters `NoGoldenMasterActionReplayResult`, which it shouldn't?

beatngu13 (changes requested):

src/main/java/de/retest/recheck/report/TestReportFilter.java
diba1013: Good point

src/test/java/de/retest/recheck/report/TestReportFilterTest.java
```diff
@@ -177,6 +194,23 @@ void test_replay_result_should_be_filtered_properly() throws Exception {
 		assertThat( differences ).containsExactly( notFilterMe );
 	}
 
+	@Test
+	void test_replay_result_should_keep_golden_master_exceptions() {
```
beatngu13: I would argue we can remove this because we have the other test now. (Otherwise, why not have the same test for other parent containers such as `SuiteReplayResult` or `TestReport`?)

beatngu13: I would argue we can remove this because we have the other test now. (Otherwise, why not have the same test for other parent containers such as `SuiteReplayResult` or `TestReport`?)

beatngu13 (changes requested):

beatngu13 (approved):

PR #385, opened 2019-08-27
Feature/improve recheck opt builder

diba1013:


pom.xml
```diff
@@ -358,6 +358,24 @@
 					<!-- downgrade to fix Maven build with Java 11 -->
 					<!-- see https://bugs.openjdk.java.net/browse/JDK-8212233 -->
 					<source>8</source>
+                    <!-- https://stackoverflow.com/questions/38707030/maven-javadoc-plugin-and-inheritdoc-for-java-api-core-classes -->
```
beatngu13: Stack Overflow has short links via "share" button:

```suggestion
                    <!-- see https://stackoverflow.com/q/38707030 -->
```

beatngu13: You seem to use whitespaces for indentation, rest of the document uses tabs—as it should be. 😉

src/main/java/de/retest/recheck/RecheckOptions.java
```diff
@@ -7,74 +7,160 @@
...
+import de.retest.recheck.persistence.FileNamer;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
-import lombok.Getter;
 
+/**
+ * This class configures the behaviour of {@link Recheck} and their implementations.
+ */
 @AllArgsConstructor( access = AccessLevel.PRIVATE )
-@Getter
```
beatngu13, line 15: IMHO you don't have to create getters manually to fix the warnings, I'm fine with them.

src/main/java/de/retest/recheck/RecheckOptions.java
```diff
@@ -7,74 +7,160 @@
...
 public class RecheckOptions {
 
 	private final FileNamerStrategy fileNamerStrategy;
 	private final String suiteName;
 	private final boolean reportUploadEnabled;
 	private final Filter filter;
+	private final List<Filter> filterToAdd;
```
beatngu13: Could you ELI5 why you moved this stuff into here? I don't quite get the commit message. Thx.

beatngu13 (changes requested):

pom.xml
diba1013: IntelliJ, always struggling with eclipse style 😉

src/main/java/de/retest/recheck/RecheckOptions.java
diba1013: I just fear, that the creation of the options and thus the call to `FileNamerStrategy#getTestClass()` might happen too early (i. e. outside the scope of a test) and thus might result in a wrong name or suite location (ignore). We cannot know, how the `FileNamerStrategy` is implemented.

Thus making sure that this code is executed in the getters ensures we are in the context of `Recheck`.

src/main/java/de/retest/recheck/RecheckOptions.java
diba1013: I did it partly, too, because I was not happy with the JavaDoc between the fields and do not know if the maven plugin conforms with lombok.

Thoughts?

beatngu13: Dunno, commenting getters somewhat feels like:

https://twitter.com/carterwickstrom/status/1014165500056596481

The code now contains a lot of noise due to the comments. But it's up to decide, you've put some effort in it.

src/main/java/de/retest/recheck/RecheckOptions.java
beatngu13: Personally, I like that `RecheckOptions` is just a data object without logic. The "complicated" part is creating the options, which is why I prefer this stuff to be inside the builder.

However, I get your point and valid arguments. I'm just trying to think of a situtation where options might be created where a test class is _not_ part of the stack trace. If someone doesn't use JUnit or TestNG, they have to provide a `FileNamerStrategy` anyway?

src/main/java/de/retest/recheck/RecheckOptions.java
diba1013: Mostly, yes (especially with `@return`). But for special getters (inside public API) as I implemented here, I think it is necessary.

src/main/java/de/retest/recheck/RecheckOptions.java
diba1013: I talked to @roesslerj&mdash;as you both reworked the options&mdash;and he approves this change, to (somehow) lazily get the correct values within `RecheckOptions`.

I agree with you that the builder should be the logic part. However, the main part is the `suiteName` which couples to the `FileNamerStrategy` and loads the filter. Thus, I think, it is more important to ensure the correct context to be sure about this name. If we can ensure that the builder is always in this context, that's fine. Maybe its just my paranoia running wild ... 😉.

beatngu13: Allright. But may I suggest to keep the single `Filter` field in the options class? This is really confusing and can be safely set via the builder.

diba1013: That's the best, I can do... We need some kind of indicator, since we have two possibilities. Either the user overwrites the filter (-> do not add the filters) or the user wants to append only filters (-> add the filters). In both cases, the builder would generate a single `Filter`.

But the default filters need the `suiteName` which was the reason, I re-did this in the first place.

beatngu13 (approved):

PR #386, opened 2019-08-27
Add methods for easily parsing filters from a string

diba1013:
to remove anoyance with constructing filters manually

### DIFF
--- a/src/main/java/de/retest/recheck/ignore/Filters.java
+++ b/src/main/java/de/retest/recheck/ignore/Filters.java
@@ -3,6 +3,7 @@ package de.retest.recheck.ignore;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


src/main/java/de/retest/recheck/ignore/Filters.java
```diff
@@ -12,10 +13,22 @@
...
-					.collect( Collectors.collectingAndThen( Collectors.toList(), CompoundFilter::new ) );
+			return parse( filterFileLines );
 		}
 	}
+
+	public static Filter parse( final String line ) {
+		return parse( Stream.of( line ) );
+	}
+
+	public static Filter parse( final Collection<String> lines ) {
```
beatngu13: `lines` are an _ordered_ collection, so I would suggest to use `List`?

beatngu13 (changes requested):

beatngu13 (approved):

PR #387, opened 2019-08-28
Fix element has attribute differences if empty due to filtering

diba1013:
When there are no differences and we try to filter or the filter removes all differences, the printer still prints verbose messages:
```
 text at foo[1]/bar[1]
     no differences
```
This resolves this issue by also checking the size of the difference.

### DIFF
--- a/src/main/java/de/retest/recheck/ui/diff/ElementDifference.java
+++ b/src/main/java/de/retest/recheck/ui/diff/ElementDifference.java
@@ -131,7 +131,7 @@ public class ElementDifference implements Difference, Comparable<ElementDifferen
 	}
 
 	public boolean hasAttributesDifferences() {
-		return attributesDifference != null;
+		return attributesDifference != null && attributesDifference.size() > 0;
 	}
 
 	public boolean hasIdentAttributesDifferences() {
--- a/src/test/java/de/retest/recheck/report/TestReportFilterTest.java
+++ b/src/test/java/de/retest/recheck/report/TestReportFilterTest.java
@@ -135,6 +135,27 @@ class TestReportFilterTest {
 	}
 
 	@Test
+	void filter_element_difference_should_have_no_differences_if_filtered() {
+		final AttributeDifference attributeDifference = mock( AttributeDifference.class );
+		final AttributesDifference attributes = mock( AttributesDifference.class );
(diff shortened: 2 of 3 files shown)


src/main/java/de/retest/recheck/ui/diff/ElementDifference.java
```diff
@@ -131,7 +131,7 @@ public static String getSumIdentifier( final Collection<ElementDifference> diffe
 	}
 
 	public boolean hasAttributesDifferences() {
-		return attributesDifference != null;
+		return attributesDifference != null && attributesDifference.size() > 0;
```
beatngu13, line 134: These `null` and `empty()` checks are getting out of control … 🤯

src/test/java/de/retest/recheck/report/TestReportFilterTest.java
```diff
@@ -134,6 +134,27 @@ void element_difference_and_child_differences_should_be_filtered_properly() thro
...
+		when( attributes.getDifferences() ).thenReturn( Collections.singletonList( attributeDifference ) );
+
+		final ElementDifference difference = mock( ElementDifference.class );
+		when( difference.getAttributesDifference() ).thenReturn( attributes );
+
+		final Filter filterAll = mock( Filter.class );
+		when( filterAll.matches( any() ) ).thenReturn( true );
+		when( filterAll.matches( any(), any() ) ).thenReturn( true );
+
+		final ElementDifference filter = TestReportFilter.filter( difference, filterAll );
```
beatngu13: `filter` is really confusing since it is an `ElementDifference`, not a filter. How about `filtered` or `filteredDifference`?

beatngu13 (changes requested):

src/main/java/de/retest/recheck/ui/diff/ElementDifference.java
diba1013: Yes, sadly that's how our code works ... . I'd prefer the latter only and have something with `AttributesDifference.empty()` but that would be a bit of a rewrite.

beatngu13: Sure, I know it's not your fault and we cannot easily change this.

beatngu13 (approved):

PR #393, opened 2019-08-30
Feature/printer optimize insert delete

diba1013:
When there are elements inserted/deleted with many children, each child prints its own insertions/deletion message. This can be simplified, if only the parent prints its message and suppresses the child differences. However, we therefore need to traverse manually through the difference tree.

### DIFF
--- a/src/main/java/de/retest/recheck/printer/ActionReplayResultPrinter.java
+++ b/src/main/java/de/retest/recheck/printer/ActionReplayResultPrinter.java
@@ -34,9 +34,6 @@ public class ActionReplayResultPrinter implements Printer<ActionReplayResult> {
 		if ( difference instanceof NoGoldenMasterActionReplayResult ) {
 			return prefix + nextIndent + NoGoldenMasterActionReplayResult.MSG_LONG;
 		}
-		if ( !difference.hasDifferences() ) {
-			return prefix + nextIndent + "no differences";
-		}
 		return prefix + createDifferences( difference.getStateDifference(), nextIndent );
 	}
 
--- a/src/main/java/de/retest/recheck/printer/ActionReplayResultPrinter.java
+++ b/src/main/java/de/retest/recheck/printer/ActionReplayResultPrinter.java
@@ -1,12 +1,15 @@
 package de.retest.recheck.printer;
 
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import de.retest.recheck.NoGoldenMasterActionReplayResult;
 import de.retest.recheck.report.ActionReplayResult;
 import de.retest.recheck.ui.DefaultValueFinder;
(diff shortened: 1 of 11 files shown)


src/test/java/de/retest/recheck/printer/ActionReplayResultPrinterTest.java
```diff
@@ -104,4 +117,66 @@ void toString_should_print_message_if_no_recheck_action_replay_result() {
...
+		final RootElementDifference rootDifference =
+				new RootElementDifference( differences, mock( RootElement.class ), mock( RootElement.class ) );
+
+		final ActionReplayResult replayResult = ActionReplayResult.withDifference( ActionReplayData.ofSutStart(),
+				WindowRetriever.empty(), DifferenceRetriever.of( Collections.singletonList( rootDifference ) ), 0L );
+
+		ApprovalsUtil.verify( cut.toString( replayResult ) );
+	}
+
+	private ElementDifference delete( final Path path, final List<ElementDifference> childDifferences ) {
```
beatngu13: Maybe signature of `delete` (and `change`) to accept both `( String path )` and `( String, List<ElementDifference> )`, then the actual test code is less verbose?

src/test/java/de/retest/recheck/printer/ActionReplayResultPrinterTest.java
```diff
@@ -118,6 +118,26 @@ void toString_should_print_message_if_no_recheck_action_replay_result() {
 		assertThat( string ).contains( NoGoldenMasterActionReplayResult.MSG_LONG );
 	}
 
+	@Test
+	void toString_should_not_print_differences_if_no_state() {
+		final ActionReplayResult replayResult = mock( ActionReplayResult.class );
+		when( replayResult.getDescription() ).thenReturn( "foo" );
+		when( replayResult.hasDifferences() ).thenCallRealMethod();
+		when( replayResult.getStateDifference() ).thenReturn( null );
```
beatngu13: AFAIK can be omitted as this is the default for mocks.

src/main/java/de/retest/recheck/printer/ActionReplayResultPrinter.java
```diff
@@ -34,16 +34,18 @@ public String toString( final ActionReplayResult difference, final String indent
 		if ( difference instanceof NoGoldenMasterActionReplayResult ) {
 			return prefix + nextIndent + NoGoldenMasterActionReplayResult.MSG_LONG;
 		}
-		return prefix + createDifferences( difference, nextIndent );
+		if ( !difference.hasDifferences() ) {
+			return prefix + nextIndent + "no differences";
```
beatngu13: Why again do we need this?

beatngu13 (changes requested):

beatngu13 (approved):

PR #397, opened 2019-09-02
Feature/zip reports

roesslerj:


src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -72,7 +78,8 @@ public void save( final URI identifier, final T element ) throws IOException {
 		final Path path = Paths.get( identifier );
 		final File file = path.toFile();
 		FileUtil.ensureFolder( path.toFile() );
-		try ( final Output output = new Output( Files.newOutputStream( path ) ) ) {
+		final OutputStream outStream = new LZ4FrameOutputStream( newOutputStream( path ) );
```
beatngu13: Shouldn't we also `try` here?

```java
try ( final Output output = new Output( new LZ4FrameOutputStream( newOutputStream( path ) ) ) {
```

beatngu13 (changes requested):

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
roesslerj: Nope. Tried ... but we need to flush and close Kryo, not LZ4...

beatngu13: Since `Output` wraps the LZ4 stream, doesn't closing it also closes LZ4? Where is the difference?

beatngu13: Since output wraps the LZ4 stream, doesn't closing it also closes LZ4? Where is the difference?

roesslerj: Difference is we want to flush the most inner stream (Output), so everything comes out. But when reading, we want to try-catch around the most outer stream.

roesslerj: Hmm. Let me try something.

roesslerj: Done

beatngu13 (approved):

PR #401, opened 2019-09-04
Use an URI, so logging tools can recognize the URI and open directly...

roesslerj:


src/main/java/de/retest/recheck/persistence/RecheckTestReportUtil.java
```diff
@@ -25,7 +24,7 @@ private RecheckTestReportUtil() {
 	}
 
 	public static void persist( final SuiteReplayResult suite, final File file ) {
-		logger.info( "Persisting test report to file '{}'.", FileUtil.canonicalPathQuietly( file ) );
+		logger.info( "Persisting test report to file '{}'.", file.getAbsoluteFile().toURI() );
```
beatngu13, line 27: Good idea, but wouldn't an URL fit better as we specifically want a location?

```suggestion
		logger.info( "Persisting test report to file '{}'.", file.getAbsoluteFile().toURL() );
```

beatngu13 (changes requested):

src/main/java/de/retest/recheck/persistence/RecheckTestReportUtil.java
roesslerj: `toURL` on file is deprecated. `toURI().toURL()` throws a checked `MalformedUrlException` and should produce the same result. Thus I would keep it as is...

beatngu13 (approved):

PR #403, opened 2019-09-06
Fix unmarshall for retest ID issue

beatngu13:
If somebody can explain this, I'm happy to hear that story …


src/test/java/de/retest/recheck/persistence/xml/XmlTransformerTest.java
```diff
@@ -0,0 +1,46 @@
...
+import de.retest.recheck.ui.descriptors.Element;
+import de.retest.recheck.ui.descriptors.SutState;
+
+class XmlTransformerTest {
+
+	@Test
+	void should_not_mix_data_from_multiple_xml_files() throws Exception {
+		final String base = "src/test/resources/de/retest/recheck/persistence/xml/";
+
+		final XmlTransformer cut = new XmlTransformer( StdXmlClassesProvider.getXmlDataClasses() );
```
martin-v: Line 155 in class XmlTransformer sets the StdXmlClasses, so it's duplicated. An empty constructor is an option.

src/test/java/de/retest/recheck/persistence/xml/XmlTransformerTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import de.retest.recheck.persistence.xml.util.StdXmlClassesProvider;
+import de.retest.recheck.ui.descriptors.Element;
+import de.retest.recheck.ui.descriptors.SutState;
+
+class XmlTransformerTest {
+
+	@Test
+	void should_not_mix_data_from_multiple_xml_files() throws Exception {
+		final String base = "src/test/resources/de/retest/recheck/persistence/xml/";
```
martin-v: Looks like a constant value, why it isn't?

martin-v (dismissed):

martin-v:
> If somebody can explain this, I'm happy to hear that story …

Looks like a bug in moxy, sadly we don't use the current version so before we file a bug we should verify it with a current version.

src/test/java/de/retest/recheck/persistence/xml/XmlTransformerTest.java
beatngu13: It is only value for this particular test method, not the entire test class, so why should I increase the scope if this thing if not necessary? However, since there are no other tests, I can introduce a constant.

roesslerj (approved):

roesslerj (approved):

src/test/resources/de/retest/recheck/persistence/xml/XmlTransformerTest.test.state0.xml
```diff
@@ -0,0 +1,17018 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
beatngu13: @Creepow will work on that afterwards (see RET-1797).

PR #406, opened 2019-09-07
We need this ctor to be protected, so subclass can call it...

roesslerj:


src/main/java/de/retest/recheck/RecheckOptions.java
```diff
@@ -26,6 +26,14 @@
 	private final Filter filter;
 	private final boolean addDefaultFilters;
 
+	public RecheckOptions( final RecheckOptions superOptions ) {
```
beatngu13: Hmmm, I'm not sure about the copy ctor—few things come into my mind:

1. It should be `protected`.
2. Why `superOptions`? At this point, the class isn't aware of this, only the copy it has to do. So I would suggest to use something like `toCopy`.
3. It is public API, so I would write a few notes. Especially, that it creates a shallow copy of the given options.
4. Reuse the existing all args ctor?

beatngu13 (changes requested):

src/main/java/de/retest/recheck/RecheckOptions.java
roesslerj: Great feedback. Done all of it.

beatngu13 (approved):

PR #408, opened 2019-09-08
Feature/improve file namer

roesslerj:
I guess it's better to review in order of commits ... or at least read the commit messages.

### DIFF
--- a/src/main/java/de/retest/recheck/FileNamerStrategy.java
+++ b/src/main/java/de/retest/recheck/FileNamerStrategy.java
@@ -17,7 +17,16 @@ public interface FileNamerStrategy {
 	 *            the base names to be used
 	 * @return a file namer using the given base names
 	 */
-	FileNamer createFileNamer( String... baseNames );
+	FileNamer createFileNamer();
+
+	/**
+	 * @deprecated Because the old FileNamer mechanism was hard to understand and implement. Use
+	 *             {@link #createFileNamer()} instead.
+	 */
+	@Deprecated
+	default FileNamer createFileNamer( final String... baseNames ) {
+		return createFileNamer();
+	}
 
 	/**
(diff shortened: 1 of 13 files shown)


diba1013:
We should make absolutely sure, that we do not break the `FileNamer` and/or `FileNamerStrategy` as @javydreamercsw mentioned several usages of the `FileNamerStrategy` per brower (version) and platform to keep the respective files separated.

I suggest creating a dummy `FileNamerStrategy` (maybe inside a test) before the changes and do not alter this for compatibility reasons.

For the `create` and `get` there are&mdash;in my opinion&mdash;two questions to consider:

1.  Who manages the lifetime or more specifically the scope of the object?

- Recheck: If `Recheck` only uses one/few instances and distributes those accordingly (thus managing its lifetime), we should call it `create`, suggesting that it is fine to directly create an instance within the method. 
- FileNamerStrategy: However, `get` suggests that the `FileNamerStrategy` is managing the lifetime of the object, suggesting that it should be created in its constructor (or similar) and be kept as a field, returning said field inside the method&mdash;thus not creating every time on method call.

2. Is there any state involved within a `FileNamer` which justifies the multiple creations? If so, I would prefer `create` otherwise `get` (assuming above point).

In either case, I suggest that we revisit the usages of these classes and choose the method name accordingly, freeing potential resources (and at best performance).

-- 

Oh and please: No interface implementations inside interfaces (besides small `@FunctionalInterface`s). This is just ugly ... besides, choose either `Path` or `File`, try not to mix them.

--

Everything else, i'm out.

roesslerj:
Great comments, thanks. I added the LegacyCompatibilityFileNamerStrategyTest ... does that suffice in your opinion?

roesslerj:
Just realized, this needs some more work...

PR #419, opened 2019-09-10
Add test from review, delete there...

roesslerj:


diba1013 (commented):
We have a lot more test that are inside review, why only this test?

roesslerj:
Because it failed when updating recheck in review. That is when I realized that this test really belongs to recheck. So I moved it when fixing it.

diba1013 (approved):

PR #425, opened 2019-09-13
Fix/img src is content

roesslerj:


diba1013:
Didn't we remove this once? Ah ... commit message.

roesslerj:
Yes. We removed it in the style filter, because we said it's not style, it's content. Consequently, we'd have to add it to the content filter...

diba1013 (approved):

PR #411, opened 2019-09-09
Add completely new interfaces and implementation for FileNamerStrategy

roesslerj:
Separation of concerns: Have NamingStrategy and ProjectLayout separated.

PR #433, opened 2019-09-18
Fix/fix custom suite name not used

diba1013:
When using the new `NamingStrategy` (introduced in #411 by @roesslerj) but provide a custom `suiteName`, it is not properly used for the golden master. Thus upon updating to the new version (from 1.4.0) with no custom `FileNamerStrategy` provided, the default mechanism jumps in and relocates the golden master. Thus a exception is thrown.

This might be a potential fix for retest/recheck-web#352.


src/test/java/de/retest/recheck/RecheckImplIT.java
```diff
@@ -29,36 +31,60 @@
...
+	}
+
+	@Test
+	void diff_should_handle_spaces_accordingly() {
+		execute( "with spaces", RecheckOptions.builder() //
+				.suiteName( RecheckImplIT.class.getName() + " spaces" ) //
+				.build() );
+	}
+
+	private RecheckOptions withIgnore( final Filter filterNothing ) {
```
beatngu13:
```suggestion
	private RecheckOptions withIgnore( final Filter ignore ) {
```

Or something like this because it isn't invoked exclusively with `Filter.FILTER_NOTHING` (see line 39).

src/test/java/de/retest/recheck/RecheckImplIT.java
```diff
@@ -29,36 +31,60 @@
...
 	@Test
 	void diff_should_be_created_if_filtered() {
-		execute( "filter", Filters.parse( "matcher: retestid=same-id" ) );
+		execute( "filter", withIgnore( Filters.parse( "matcher: retestid=same-id" ) ) );
 	}
 
-	void execute( final String name, final Filter filter ) {
+	@Test
+	void diff_should_handle_legacy_spaces_accordingly() {
+		final MavenConformFileNamerStrategy fileNamerStrategy = spy( new MavenConformFileNamerStrategy() );
```
beatngu13: Use interface:

```suggestion
		final FileNamerStrategy fileNamerStrategy = spy( new MavenConformFileNamerStrategy() );
```

src/test/java/de/retest/recheck/RecheckImplIT.java
```diff
@@ -79,9 +79,12 @@ void execute( final String name, final RootElement toVerify, final RecheckOption
 
 		re.startTest( name );
 		re.check( toVerify, new RootElementAdapter(), "check" );
-		re.capTest();
 
-		re.cap();
+		try {
+			re.capTest();
+		} finally {
+			re.cap();
```
beatngu13, line 86: Why not in `@AfterEach`?

beatngu13 (changes requested):

src/test/java/de/retest/recheck/RecheckImplIT.java
diba1013: I emulate the lifecycle of Recheck within a test twice so that a error message can be produced, thus this needs to be called twice (as we did not update the lifecycle yet).

beatngu13 (approved):

PR #434, opened 2019-09-18
Fix naming

beatngu13:
See commit message.


roesslerj:
Hmm. I just yesterday had a conversation with someone for which this doesn't work, as he uses a `CucumberWithSerenity` runner. How would you name that in contrast then?

beatngu13:
Then this is caused by lacking support from our test class and method namer, isn't it? I mean, what the class does is now reflected in the class name: it names suites and tests according to class and method names. This feature supports out-of-the-box several frameworks, currently JUnit and TestNG. Naming it "JUnit-something" and having support for TestNG is more confusing I believe.

rebazer:

rebazer:

roesslerj (approved):

PR #435, opened 2019-09-18
Fix filtering of inserted and deleted elements

beatngu13:


src/main/java/de/retest/recheck/report/TestReportFilter.java
```diff
@@ -130,6 +138,13 @@ static IdentifyingAttributesDifference filter( final Element element,
 						: new IdentifyingAttributesDifference( element.getIdentifyingAttributes(), diffs ) ) );
 	}
 
+	static InsertedDeletedElementDifference filter( final InsertedDeletedElementDifference insertedDeletedDiff,
+			final Filter filter ) {
+		final Element insertedDeleted =
+				insertedDeletedDiff.isInserted() ? insertedDeletedDiff.getActual() : insertedDeletedDiff.getExpected();
+		return filter.matches( insertedDeleted ) ? null : insertedDeletedDiff;
```
diba1013: Now we call `filter.matches(Element)` but I still think, that this needs to happen for general elements too.

src/main/java/de/retest/recheck/report/TestReportFilter.java
```diff
@@ -96,16 +97,23 @@ static StateDifference filter( final StateDifference stateDifference, final Filt
 		AttributesDifference attributesDifference = elementDiff.getAttributesDifference();
 		LeafDifference identifyingAttributesDifference = elementDiff.getIdentifyingAttributesDifference();
 		Collection<ElementDifference> childDifferences = elementDiff.getChildDifferences();
+
 		if ( elementDiff.hasAttributesDifferences() ) {
 			attributesDifference = filter( elementDiff.getElement(), elementDiff.getAttributesDifference(), filter );
 		}
+
 		if ( elementDiff.hasIdentAttributesDifferences() ) {
```
diba1013, line 105: Does it make sense, that we move this into a separate method:
```
LeafDifference identifyingAttributesDifference = filterIdentifyingAttributes(elementDiff, filter)
```

src/main/java/de/retest/recheck/report/TestReportFilter.java
```diff
@@ -130,6 +138,13 @@ static IdentifyingAttributesDifference filter( final Element element,
 						: new IdentifyingAttributesDifference( element.getIdentifyingAttributes(), diffs ) ) );
 	}
 
+	static InsertedDeletedElementDifference filter( final InsertedDeletedElementDifference insertedDeletedDiff,
+			final Filter filter ) {
+		final Element insertedDeleted =
+				insertedDeletedDiff.isInserted() ? insertedDeletedDiff.getActual() : insertedDeletedDiff.getExpected();
```
diba1013: Should we move this into `InsertedDeletedElementDifference`. I assume, we need this a couple more times...

diba1013 (changes requested):

src/main/java/de/retest/recheck/report/TestReportFilter.java
beatngu13: I believe it does work because `Filter#matches( Element, AttributeDifference )` defaults to `Filter#matches( Element )` if a filter only implements the latter. See the following assert: https://github.com/retest/recheck/blob/2e6fe13e88b514aae49331ab52f940bdea2ccd00/src/test/java/de/retest/recheck/report/TestReportFilterTest.java#L210

src/main/java/de/retest/recheck/report/TestReportFilter.java
beatngu13: IMHO it is OK to stick to the given style.

diba1013 (approved):

PR #438, opened 2019-09-20
Feature/fix pixel diff filter

enesoezel:


src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
```diff
@@ -15,6 +18,7 @@
 @Getter
 @Slf4j
 public class PixelDiffFilter implements Filter {
+	//TODO Move this web specific class to recheck-web
```
beatngu13: As it belongs to the class, I suggest to move it above the annotations. (And please insert a space: `// TODO ...` 🤓)

src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
```diff
@@ -25,6 +29,8 @@
 	private final boolean specifiedAsDouble;
 	private final double pixelDiff;
 
+	private static final Set<String> ignoredKeys = new HashSet<>( Collections.singletonList( "style" ) );
```
beatngu13: Should be moved above the instance fields, where also the other `static` field is.

src/test/java/de/retest/recheck/review/ignore/PixelDiffFilterTest.java
```diff
@@ -87,4 +87,13 @@ void specifiedAsDouble_flag_should_control_toString() throws Exception {
 		assertThat( new PixelDiffFilter( true, 0.0 ) ).hasToString( "pixel-diff=0.0" );
 	}
 
+	@Test
+	void style_attribute_should_return_false() {
+		final String expected = "18px";
```
beatngu13: You could use "real-world" test data:

```java
final String expected = "width: 1px; height: 1px";
final String actual = "width: 1000px; height: 1000px";
```

beatngu13 (changes requested):
Could please also adapt the log? I don't think we need the `NumberFormatException` in there as it is quite ugly and we know what the issue is. Also it would be helpful to quote expected and actual:

```java
log.error( "Could not parse expected '{}' and actual '{}' for pixel diff.", expected, actual );
```

src/main/java/de/retest/recheck/review/ignore/PixelDiffFilter.java
diba1013: For now?
```suggestion
	private static final Set<String> ignoredKeys = Collections.singleton( "style" );
```

beatngu13 (approved):

PR #444, opened 2019-09-26
Feature/find test class name

Creepow:


src/main/java/de/retest/recheck/TestCaseFinder.java
```diff
@@ -86,7 +88,41 @@ public static TestCaseFinder getInstance() {
...
+		return Optional.empty();
+	}
+
+	/**
+	 * @param trace
+	 *            The trace to be used for search.
+	 * @return The class name for the test case method in the given stack trace.
+	 */
+	public Optional<String> findTestCaseClassInStack( final StackTraceElement[] trace ) {
+		for ( final StackTraceElement element : trace ) {
```
beatngu13: You could simplify this and the other loop using a stream:

```java
return Arrays.stream( trace ) //
		.map( StackTraceElement::getClassName ) //
		.filter( testClassSuffixes::contains ) //
		.findFirst();
```

beatngu13 (changes requested):
Could you please also update `CHANGELOG.md`?

src/main/java/de/retest/recheck/TestCaseFinder.java
```diff
@@ -86,7 +88,41 @@ public static TestCaseFinder getInstance() {
...
+		final Optional<String> testCaseClassName = findTestCaseMethodInStack( toClassName() );
+		if ( testCaseClassName.isPresent() ) {
+			return testCaseClassName;
+		} else {
+			return findTestCaseClassInStack();
+		}
+	}
+
+	/**
+	 * @return The class name for the test case method in all stack traces.
```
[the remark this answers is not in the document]
Creepow: copy & pase yes but thats the sentence, that's on almost every other method, including those who track for the class name, who already existed. So i kept it out of consistency. Mayle take a look at the other comments and tell me if they should be changed too or if i'm wrong.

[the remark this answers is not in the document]
beatngu13: Yes, the other methods work on the basis of test case methods (respectively their annotations). But this method doesn't, so that's why I suggested the changes here and below.

src/main/java/de/retest/recheck/TestCaseFinder.java
```diff
@@ -86,7 +88,41 @@ public static TestCaseFinder getInstance() {
...
+
+	/**
+	 * @param trace
+	 *            The trace to be used for search.
+	 * @return The class name for the test case method in the given stack trace.
+	 */
+	public Optional<String> findTestCaseClassInStack( final StackTraceElement[] trace ) {
+		for ( final StackTraceElement element : trace ) {
+			final String className = element.getClassName();
+			if ( StringUtils.endsWith( className, "Test" ) || StringUtils.endsWith( className, "IT" ) ) {
```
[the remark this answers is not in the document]
diba1013: If we do it so, we might want to support all `*Test*`-regex types of these plugins and not just a basic `equals`

src/main/java/de/retest/recheck/TestCaseFinder.java
```diff
@@ -86,7 +88,41 @@ public static TestCaseFinder getInstance() {
...
+		} else {
+			return findTestCaseClassInStack();
+		}
+	}
+
+	/**
+	 * @return The class name for the test being called from.
+	 */
+	public Optional<String> findTestCaseClassInStack() {
+		for ( final StackTraceElement[] stack : Thread.getAllStackTraces().values() ) {
```
diba1013: This removes the need for two methods:
```
return Thread.getAllStackTraces().values().stream() //
				.flatMap( Stream::of ) //
				.map( StackTraceElement::getClassName ) //
				.filter( testClassSuffixes::contains ) //
				.findFirst();
```

src/main/java/de/retest/recheck/TestCaseFinder.java
```diff
@@ -86,7 +88,41 @@ public static TestCaseFinder getInstance() {
...
+		}
+		return Optional.empty();
+	}
+
+	/**
+	 * @param trace
+	 *            The trace to be used for search.
+	 * @return The class name for the test being called from.
+	 */
+	public Optional<String> findTestCaseClassInStack( final StackTraceElement[] trace ) {
```
diba1013, line 128: If you keep this method, I would suggest to keep this private.

src/main/java/de/retest/recheck/TestCaseFinder.java
```diff
@@ -86,7 +88,41 @@ public static TestCaseFinder getInstance() {
...
+			return testCaseClassName;
+		} else {
+			return findTestCaseClassInStack();
+		}
+	}
+
+	/**
+	 * @return The class name for the test being called from.
+	 */
+	public Optional<String> findTestCaseClassInStack() {
```
diba1013, line 116: Does this have to be public? If yes, I would add more documentation on what this method does, i.e. what `*Test` classes does it find. If no, I still would add some more doc to the above method.

src/main/java/de/retest/recheck/TestCaseFinder.java
```diff
@@ -86,7 +88,41 @@ public static TestCaseFinder getInstance() {
 	 * @return The class name for the test case method in all stack traces.
 	 */
 	public Optional<String> findTestCaseClassNameInStack() {
-		return findTestCaseMethodInStack( toClassName() );
+		final Optional<String> testCaseClassName = findTestCaseMethodInStack( toClassName() );
+		if ( testCaseClassName.isPresent() ) {
+			return testCaseClassName;
+		} else {
+			return findTestCaseClassInStack();
+		}
```
diba1013, line 110: Now we traverse the stack trace twice (in the worst case). This is an expensive operation, that's why they introduced the Stack-Walking API in 9.

Does it make sense to unify these methods?

src/main/java/de/retest/recheck/TestCaseFinder.java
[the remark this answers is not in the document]
diba1013: If we do it so, we might want to support all `*Test*`-regex types of these plugins and not just a basic `equals`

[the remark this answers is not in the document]
diba1013: Also, do we have any information about how TestNG does its test (since we claim to support it).

diba1013 (dismissed):

src/main/java/de/retest/recheck/TestCaseFinder.java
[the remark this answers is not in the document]
beatngu13: For [Surefire](https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html):

>By default, the Surefire Plugin will automatically include all test classes with the following wildcard patterns:
>
>* `"**/Test*.java"` - includes all of its subdirectories and all Java filenames that start with `"Test".
>* `"**/*Test.java"` - includes all of its subdirectories and all Java filenames that end with `"Test".
>* `"**/*Tests.java"` - includes all of its subdirectories and all Java filenames that end with "Tests".
>* `"**/*TestCase.java"` - includes all of its subdirectories and all Java filenames that end with "TestCase".

For [Failsafe](https://maven.apache.org/surefire/maven-failsafe-plugin/examples/inclusion-exclusion.html):

>By default, the Failsafe Plugin will automatically include all test classes with the following wildcard patterns:
>
>* `"**/IT*.java"` - includes all of its subdirectories and all Java filenames that start with "IT".
>* `"**/*IT.java"` - includes all of its subdirectories and all Java filenames that end with "IT".
>* `"**/*ITCase.java"` - includes all of its subdirectories and all Java filenames that end with "ITCase".exclusion.html

For [TestNG](https://testng.org/doc/documentation-main.html):

>A TestNG class is a Java class that contains at least one TestNG annotation.

So I assume names don't matter here, just like with JUnit.

src/main/java/de/retest/recheck/TestCaseFinder.java
beatngu13: Pascal already incorporated a stream-based approach. However, we decided to have public variants with and without passing a stack trace, mostly for being consistent with the remaining API.

src/main/java/de/retest/recheck/TestCaseFinder.java
beatngu13: There is public `findTestCaseMethodInStack( StackTraceElement[] )`.

src/main/java/de/retest/recheck/TestCaseFinder.java
beatngu13: There is public `findTestCaseMethodInStack()`.

src/main/java/de/retest/recheck/TestCaseFinder.java
beatngu13: There is public `findTestCaseMethodInStack()`.

src/main/java/de/retest/recheck/TestCaseFinder.java
beatngu13: I thought about this too, but this would require some refactoring. I'm really not sure if this is worth the effort …

src/main/java/de/retest/recheck/TestCaseFinder.java
```diff
@@ -51,6 +53,9 @@
 			"org.junit.jupiter.api.RepeatedTest", //
 			"org.junit.jupiter.params.ParameterizedTest" ) );
 
+	private static final List<String> suffixes = Arrays.asList( "Test", "IT", "TestCase", "ITCase", "Tests" );
```
beatngu13: "suffixes" and "prefixes" of what? 🤷‍♂️ Just like with `testCaseAnnotations`, it may become more clear if you do something like:

```java
private static final List<String> testClassSuffixes = Arrays.asList( //
        // Maven Surefire
        "Test",
        // ...
        // Maven Failsaife
        "IT"
         // ...
```

I suggest to do the same for the prefixes.

beatngu13 (changes requested):

beatngu13 (approved):

PR #445, opened 2019-09-27
Replace okhttp with unirest

modulo11:
The current version (4.x) of okhttp is shipped as a Kotlin library. While this generally is a good thing, it blows up some internal parts of recheck and recheck-web due to dependency issues. As we do not heavily rely on okhttp it was and will be pretty easy to replace it with unirest.

### DIFF
--- a/src/main/java/de/retest/recheck/auth/RetestAuthentication.java
+++ b/src/main/java/de/retest/recheck/auth/RetestAuthentication.java
@@ -10,8 +10,13 @@ import java.io.PrintWriter;
 import java.net.ServerSocket;
 import java.net.Socket;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
+import org.apache.http.NameValuePair;
+import org.apache.http.client.utils.URLEncodedUtils;
 import org.keycloak.OAuth2Constants;
 import org.keycloak.OAuthErrorException;
 import org.keycloak.adapters.KeycloakDeployment;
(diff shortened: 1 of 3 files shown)


src/main/java/de/retest/recheck/auth/RetestAuthentication.java
```diff
@@ -177,13 +181,16 @@ private boolean isTokenValid() {
 	}
 
 	static KeycloakResult getRequestParameters( final String request ) {
-		final HttpUrl url = HttpUrl.parse( "http://localhost/" ).resolve( request.split( " " )[1] );
+		final String url = "http://localhost/" + request.split( " " )[1];
+		final Map<String, String> parameters = URLEncodedUtils.parse( URI.create( url ), Charset.forName( "UTF-8" ) )//
```
beatngu13:
```suggestion
		final Map<String, String> parameters = URLEncodedUtils.parse( URI.create( url ), StandardCharsets.UTF8 ) //
```

src/main/java/de/retest/recheck/auth/RetestAuthentication.java
```diff
@@ -177,13 +181,16 @@ private boolean isTokenValid() {
...
+				.stream()//
```
beatngu13: Sometimes you do `...)//` (e.g. here and line 193) and sometimes `...) //`.

beatngu13 (changes requested):

src/main/java/de/retest/recheck/auth/RetestAuthentication.java
modulo11: :see_no_evil:

beatngu13 (approved):

PR #455, opened 2019-10-09
Feature/configurable window match

roesslerj:


src/main/java/de/retest/recheck/Properties.java
```diff
@@ -1,7 +1,12 @@
 package de.retest.recheck;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class Properties {
 
+	private static final Logger logger = LoggerFactory.getLogger( Properties.class );
```
diba1013: Use `@Slf4j` from lombok.

src/main/java/de/retest/recheck/Properties.java
```diff
@@ -49,4 +57,18 @@ public static FileOutputFormat getStateOutputFormat() {
 		}
 		return FileOutputFormat.PLAIN;
 	}
+
+	public static Double getConfiguredDouble( final String property, final Double defaultValue ) {
```
diba1013: Use primitives `double`

src/main/java/de/retest/recheck/ui/diff/Alignment.java
```diff
@@ -23,7 +25,8 @@
 	private final Map<Element, Element> alignment;
 	private final Map<Element, Element> expectedMapOfElementTree = new HashMap<>();
 	private final Map<Element, Element> actualMapOfElementTree = new HashMap<>();
-	private final Double minimumMatch = getRequiredMinimumMatch();
+	private final Double minimumMatch =
+			getConfiguredDouble( ELEMENT_MATCH_THRESHOLD_PROPERTY, ELEMENT_MATCH_THRESHOLD_DEFAULT );
```
diba1013: I would move this method into Properties as `getElementMatchThreshhold()` and `getWindowMatchThreshold()`.

src/main/java/de/retest/recheck/Properties.java
```diff
@@ -10,8 +15,11 @@
 	public static final String PROPERTY_VALUE_SEPARATOR = ";";
 	public static final String ZIP_FOLDER_SEPARATOR = "/";
 
-	public static final double WINDOW_MATCH_THRESHOLD = 0.8;
-	public static final double WINDOW_CHILDS_MATCH_THRESHOLD = 0.5;
+	public static final double WINDOW_MATCH_THRESHOLD_DEFAULT = 0.8;
+	public static final String WINDOW_MATCH_THRESHOLD_PROPERTY = "de.retest.recheck.windowMatchThreshold";
```
diba1013: We do not have a window per se. Should we rename this to `ROOT_ELEMENT_THRESHOLD_PROPERTY`?

src/main/java/de/retest/recheck/ui/diff/Alignment.java
diba1013: I would inline this so that it is evaluated every time it is needed (in case it is not set at construction of this class). Also this is only used once.

src/main/java/de/retest/recheck/ui/diff/RootElementDifferenceFinder.java
```diff
@@ -18,6 +21,11 @@
 
 	private final ElementDifferenceFinder elementDifferenceFinder;
 
+	private final Double minimumWindowMatch =
```
diba1013: I would inline this so that it is evaluated every time it is needed (in case it is not set at construction of this class). Also this is only used once.

src/main/java/de/retest/recheck/ui/diff/RootElementDifferenceFinder.java
```diff
@@ -18,6 +21,11 @@
...
+			getConfiguredDouble( WINDOW_MATCH_THRESHOLD_PROPERTY, WINDOW_MATCH_THRESHOLD_DEFAULT );
+	private final Double minimumContainedComponentsMatchMatch = getConfiguredDouble(
```
diba1013: I would inline this so that it is evaluated every time it is needed (in case it is not set at construction of this class). Also this is only used once.

diba1013 (changes requested):
I would really strive to move all system properties into `RecheckOptions`, so that they are easily accessible and one does not have to fiddle around with resetting the property within tests.

**Documentation**?
**Changelog**?
or only internal use?

src/main/java/de/retest/recheck/Properties.java
```diff
@@ -49,4 +57,18 @@ public static FileOutputFormat getStateOutputFormat() {
...
+		if ( value == null ) {
+			return defaultValue;
+		}
+		try {
+			return Double.parseDouble( value );
+		} catch ( final NumberFormatException e ) {
+			logger.error( "Exception parsing value {} of property {} to double, using default {} instead.", value,
+					property, defaultValue );
+			return defaultValue;
+		}
```
diba1013: `NumberUtils.toDouble`

roesslerj: No, actually we want an error logged, not silently the default value used...

src/main/java/de/retest/recheck/Properties.java
roesslerj: Done

src/main/java/de/retest/recheck/Properties.java
roesslerj: Done

src/main/java/de/retest/recheck/Properties.java
roesslerj: Done

src/main/java/de/retest/recheck/ui/diff/Alignment.java
roesslerj: Ok.

src/main/java/de/retest/recheck/ui/diff/RootElementDifferenceFinder.java
roesslerj: I don't think this is necessary.

src/main/java/de/retest/recheck/ui/diff/RootElementDifferenceFinder.java
roesslerj: I don't think this is necessary.

src/main/java/de/retest/recheck/ui/diff/Alignment.java
roesslerj: I don't think this is necessary.

roesslerj (commented):
This is only for internal use for now, so I wouldn't move this to Options.

diba1013 (approved):

PR #456, opened 2019-10-09
A filter.js file should be loaded as JavaScript filter

roesslerj:


diba1013 (changes requested):
I assume that the real problem is in [`SearchFilterFiles`](https://github.com/retest/recheck/blob/bfcbd2da2209e2db5f63a4dbb629c4fe1ed788f8/src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java#L101).

I really need to refactor this stuff to have `FilterProviders` so that we do not hack around this annoying `JSFilterImpl`.

As for now, I cannot really decide if `Filters` should only load text based filters with our syntax. Then you would have to move this code into `FilterLoader`. What do you think? _As always, this is the quick and dirty solution._

roesslerj:
Hmm. From what I understand, this is the (currently) best and correct solution. I wouldn't call it quick and dirty. If you want something changed, please be more specific...

diba1013:
> If you want something changed, please be more specific

Sorry, I thought I was. 

This bug was introduced, since we want custom `.filter.js` to be loaded (since the ignore.filter.js does not suffice). Thus we start in `SearchFilterFiles`. If we follow the control flow, we get to `FilterLoader#load` which accepts all filters as path (both the `.filter` and `.filter.js`). This is, so that we do not excessive IO every time we want to access the filter by name. But the `FilterLoader` essentially delegates it to `Filters#load`.

This means we could change it at two places:

1. The `FilterLoader` knows, how to load an appropriate Filter. This means it can distinguish between Text and JavaScript and thus loads them correctly. This would  somewhat keep the loading into the context of `SearchFilterFiles`, while `Filters` keeps to our syntax.
2. The `Filters#load` loads the filter and does the filter (as suggested here). This would transform the `Filters` into a _Knows-it-all-Class_ and keep the `FilterLoader` just as a lazy loader.

**I do not have a strong preference**, as we definitely need to rewrite this part of the filter searching/providing. I would be fine with the current solution. **I just want your opinion.**

roesslerj:
Thanks for making it much clearer. I moved the implementation to FilterLoader where, according to your explanation, it makes much more sense.

diba1013 (approved):

PR #457, opened 2019-10-10
Use Path inside FilterNotFoundException

diba1013:
To my understanding, this should not break any projects downstream, as this was introduced down wile the release for 1.5.0 was ongoing, thus not on the release branch.


roesslerj (approved):

diba1013:
Sonarcloud has some issues with serializable. Should we address this or could you disable this message?

roesslerj:
I would follow this rule. Just have a String instead of the Path object...

diba1013:
Since we want to use serializable, this PR is pointless as Path (or its Subclasses) are not serializable.

PR #459, opened 2019-10-10
Feature/negative duration diff

Creepow:
Delete Duration Difference


src/main/java/de/retest/recheck/util/DurationUtil.java
```diff
@@ -28,7 +28,7 @@ public static String durationToFormattedString( final long duration ) {
 	}
```
diba1013: Couldn't you delete this class and associated tests (like see below)?

diba1013 (changes requested):
Please make sure that there are no Duration* related files (e.g. Approvals) around.

diba1013 (approved):

PR #460, opened 2019-10-10
Element class matcher

beatngu13:
Before approve, let's discuss something tomorrow.

### DIFF
--- a/src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
+++ b/src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
@@ -1,0 +1,52 @@
+package de.retest.recheck.review.ignore.matcher;
+
+import java.util.Objects;
+import java.util.regex.MatchResult;
+import java.util.regex.Pattern;
+
+import de.retest.recheck.review.ignore.io.RegexLoader;
+import de.retest.recheck.ui.descriptors.Element;
+
+public class ElementClassMatcher implements Matcher<Element> {
+
+	public static final String CLASS_KEY = "class";
+
+	private final String classValue;
+
+	public ElementClassMatcher( final Element element ) {
+		this( Objects.toString( element.getIdentifyingAttributes().get( CLASS_KEY ), "" ) );
+	}
+
+	private ElementClassMatcher( final String classValue ) {
+		this.classValue = classValue;
+	}
+
+	@Override
+	public boolean test( final Element element ) {
+		final String classValue = element.getIdentifyingAttributes().get( CLASS_KEY );
+		return classValue != null && classValue.matches( this.classValue );
+	}
+
+	@Override
+	public String toString() {
+		return String.format( ElementClassMatcherLoader.FORMAT, classValue );
+	}
+
+	public static final class ElementClassMatcherLoader extends RegexLoader<ElementClassMatcher> {
+
+		private static final String CLASS = CLASS_KEY + "=";
+
+		private static final String FORMAT = CLASS + "%s";
+		private static final Pattern REGEX = Pattern.compile( CLASS + "(.+)" );
+
+		public ElementClassMatcherLoader() {
+			super( REGEX );
+		}
+
+		@Override
+		protected ElementClassMatcher load( final MatchResult matcher ) {
+			final String classValue = matcher.group( 1 );
+			return new ElementClassMatcher( classValue );
+		}
+	}
+}
--- a/src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
+++ b/src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
@@ -1,8 +1,13 @@
 package de.retest.recheck.review.ignore.matcher;
 
-import java.util.Objects;
+import static java.util.stream.Collectors.toCollection;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.regex.MatchResult;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import de.retest.recheck.review.ignore.io.RegexLoader;
 import de.retest.recheck.ui.descriptors.Element;
@@ -11,25 +16,28 @@ public class ElementClassMatcher implements Matcher<Element> {
 
 	public static final String CLASS_KEY = "class";
 
-	private final String classValue;
+	private final Set<String> classValues;
 
 	public ElementClassMatcher( final Element element ) {
-		this( Objects.toString( element.getIdentifyingAttributes().get( CLASS_KEY ), "" ) );
+		this( toClassValuesSet( toClassValueString( element ) ) );
 	}
 
-	private ElementClassMatcher( final String classValue ) {
-		this.classValue = classValue;
+	private ElementClassMatcher( final Set<String> classValues ) {
+		this.classValues = classValues;
 	}
 
 	@Override
 	public boolean test( final Element element ) {
-		final String classValue = element.getIdentifyingAttributes().get( CLASS_KEY );
-		return classValue != null && this.classValue.contains( classValue );
+		final Set<String> classValues = toClassValuesSet( toClassValueString( element ) );
+		if ( classValues.isEmpty() ) {
+			return false;
+		}
+		return this.classValues.containsAll( classValues );
 	}
 
 	@Override
 	public String toString() {
-		return String.format( ElementClassMatcherLoader.FORMAT, classValue );
+		return String.format( ElementClassMatcherLoader.FORMAT, classValues );
 	}
 
 	public static final class ElementClassMatcherLoader extends RegexLoader<ElementClassMatcher> {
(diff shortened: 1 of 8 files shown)


src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
```diff
@@ -0,0 +1,52 @@
...
+	public String toString() {
+		return String.format( ElementClassMatcherLoader.FORMAT, classValue );
+	}
+
+	public static final class ElementClassMatcherLoader extends RegexLoader<ElementClassMatcher> {
+
+		private static final String CLASS = CLASS_KEY + "=";
+
+		private static final String FORMAT = CLASS + "%s";
+		private static final Pattern REGEX = Pattern.compile( CLASS + "(.+)" );
```
diba1013, line 47: What do we do when one wants to address multiple classes, as one single class rarely fits?

```text
# with the above changes, this technically is not equal
matcher: class=flash-error
# do we want a separate syntax for that (i.e. contains)?
matcher: class~=flash 
# escape multiple with spaces, if so, when do we care about order?
matcher: class="flash-error flash"
matcher: class~="flash-error flash"
```

```
<div id="ajax-error-message" class="ajax-error-message flash flash-error"></div>
```

src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
```diff
@@ -0,0 +1,52 @@
+package de.retest.recheck.review.ignore.matcher;
+
+import java.util.Objects;
+import java.util.regex.MatchResult;
+import java.util.regex.Pattern;
+
+import de.retest.recheck.review.ignore.io.RegexLoader;
+import de.retest.recheck.ui.descriptors.Element;
+
+public class ElementClassMatcher implements Matcher<Element> {
```
diba1013, line 13: Another web based filter in recheck 😄

diba1013 (changes requested):
Classes are not an good identifying attribute, as the are supposed to be used by multiple elements. When working with the [docs test](https://github.com/diba1013/docs-test/blob/master/src/test/java/de/retest/docs/pages/SearchElement.java), I find myself needing more precise methods of identifying an element, which&mdash;as I have no real control over the html code generated&mdash;admittedly goes back to using classes inside a XPath for the above mentioned precision.

Thus for the future, we ultimately need more than one identifier to lookup an element. The more I think about a new filter syntax, the more am I drawn to something like XPath, which is designed to work for a tree like structure.

For the documentation, I think (especially now with classes) we need to emphasize, that while a single line might represent a filter, the filter itself (i.e. the matcher for an element) might match multiple elements.

Also worth mentioning: We now have multiple matchers that try to query the identifying attributes. Why not make a generic one that looks for `matcher: key=value` inside the IdentifyingAttributes? This would work for classes too, if we do not want to provide below convenience.

src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	private ElementClassMatcher( final String classValue ) {
+		this.classValue = classValue;
+	}
+
+	@Override
+	public boolean test( final Element element ) {
+		final String classValue = element.getIdentifyingAttributes().get( CLASS_KEY );
+		return classValue != null && classValue.matches( this.classValue );
```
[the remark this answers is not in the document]
beatngu13: Not relevant as it only appears once (see above)?

src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
beatngu13: Not relevant as it only appears once (see above)?

src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
beatngu13: Yep, but this is a separate issue. 😉

beatngu13:
>Classes are not an good identifying attribute, as the are supposed to be used by multiple elements. When working with the docs test, I find myself needing more precise methods of identifying an element, which—as I have no real control over the html code generated—admittedly goes back to using classes inside a XPath for the above mentioned precision.

Well, this may be true for the use case you describe here—and this totally legitimate—but I faced a different situation: I wanted to ignore image carousels on a website and they all shared the same class. But since we don't have a matcher for this, I had to select them by XPath, which easily breaks. Also note that Selenium facilities for this: https://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/By.html#className-java.lang.String-

>Thus for the future, we ultimately need more than one identifier to lookup an element. The more I think about a new filter syntax, the more am I drawn to something like XPath, which is designed to work for a tree like structure.

XPath is one option, but I would say that all methods defined in Selenium's `By` are sensible.

>For the documentation, I think (especially now with classes) we need to emphasize, that while a single line might represent a filter, the filter itself (i.e. the matcher for an element) might match multiple elements.

That's what I wanted to discuss with you. The Selenium Javadoc for `By.ByClassName` doesn't define what happens if multiple elements match the criteria. Can we guarantee a search/find order? Not sure about this.

>Also worth mentioning: We now have multiple matchers that try to query the identifying attributes. Why not make a generic one that looks for matcher: key=value inside the IdentifyingAttributes? This would work for classes too, if we do not want to provide below convenience.

I like the idea, but I'm not sure if this should be part of this PR, or if this is more of separate refactoring task.

src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
```diff
@@ -0,0 +1,52 @@
...
+import de.retest.recheck.ui.descriptors.Element;
+
+public class ElementClassMatcher implements Matcher<Element> {
+
+	public static final String CLASS_KEY = "class";
+
+	private final String classValue;
+
+	public ElementClassMatcher( final Element element ) {
+		this( Objects.toString( element.getIdentifyingAttributes().get( CLASS_KEY ), "" ) );
```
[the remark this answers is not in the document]
diba1013: That is what I assumed.

src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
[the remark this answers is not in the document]
diba1013: I am aware that our Attributes essentially represent a `Map<String, Object`>. I built upon the assumption of one big string, where it essentially represents a sequence of classes, concatenated by a space (as described above).

Thus, if we want to find a single class within the sequence, we need to process it as such a `Set<String>`.

Take Seleniums `By.className` for example:
> Find elements based on the value of the "class" attribute. If an element has multiple classes, then this will match against each of them. For example, if the value is "one two onone", then the class names "one" and "two" will match.

I think that we need to match this behavior.

diba1013:
> I had to select them by XPath, which easily break

You do not have to define a XPath by the full relative path (by XPath specification, which we currently not support). You may choose some [loose](https://github.com/diba1013/docs-test/blob/a49b21bb0dca30db0de62a77396dce5685f3fb52/src/test/java/de/retest/docs/pages/HeaderNavSourceElement.java#L19) path that still matches the and only the element you want to address. 

> XPath is one option, but I would say that all methods defined in Selenium's By are sensible.

Sure thing, but as soon as you want to combine multiple `By`s&mdash;as [these](https://github.com/diba1013/docs-test/blob/master/src/test/java/de/retest/docs/pages/SearchElement.java) examples show&mdash;you will easily find yourself not satisfied with the other Selenium `By`s. That is also a problem with our current filter syntax. Funnily enough, Selenium itself falls back to using XPath, if the `By`s cannot be satisfied.

> Can we guarantee a search/find order?

The `Filter` does not search for anything, so does the `Matcher`. I just checks if a single element satisfies some criteria. We could clarify this in the documentation, especially in context of stateless. I initially thought of the same issue, that it must use `Matcher<List<Element>`, but it looks at the current (context) element, not its children.

src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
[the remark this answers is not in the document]
beatngu13: I thought `matches` handles this, I now use `contains`. What do you think?

beatngu13:
>You do not have to define a XPath by the full relative path (by XPath specification, which we currently not support). You may choose some loose path that still matches the and only the element you want to address.

I'm talking about our filter syntax, which only supports absolute XPaths. 😕

Let's talk about the remaining stuff via Slack.

src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
beatngu13: Adapted the implementation accordingly, still open to discuss:

```
matcher: class=one two
# or
matcher: class="one two"
```

And one more thing: The `Set` currently doesn't guarantee order, so if the above is loaded, it may be saved as "two one"—which we have to handle.

beatngu13: So I've adapted the things we have discussed, but I'm not sure if we should require the quoted variant over the unquoted one. Consider:

```
# should work
matcher: class=one
matcher: class="one"
matcher: class="one two"

# should not work
matcher: class=one two
```

This really complicates the implementation and tests, and there are several ways to do it. To come up with a proper solution, I'd suggest to further discuss and possibly pair this on Monday.

beatngu13:
Adapted the branch as discussed, here's the corresponding documentation: https://github.com/retest/docs/pull/54

diba1013 (approved):

PR #462, opened 2019-10-11
Move retestId provider from Properties to Options

roesslerj:


diba1013:
This is currently not used with in your PR! What is the default `RetestIdProvider`? If you did not mean to open a draft, please provide a proper description on what your intention is with this PR, as to me it seems to be in an unfinished state.

Regardless, I think we need to do a workshop before, where we discuss how to get rid of the system properties. I would even argue, to not do this now and wait for 2.0 or a major rewrite where we do it properly.

As such, I would advise for now: If you want it to be configurable by the user, document it inside the [docs](https://docs.retest.de/recheck/usage/configuration/) under `Properties`.

This should be discussed:

1. I would advise against for `Recheck` to just set the system property as this is just a ugly communication workaround (i.e. set the class name and recreate it every time we want to access it).
2. Then how would one get access to the `RetestIdProvider`? Do we inject those (along with a couple other fields) into a `RecheckAdapter`?
3. `Recheck` as such does not access the `RecheckOption` or creates another data container (e.g. `RecheckContext`) and just pass it into the algorithm which extracts the needed options?
4. Use DI?

PR #463, opened 2019-10-12
Feature/move props to opts

roesslerj:
I oppose to wait for a 2.0 that will cause a big rewriting, introduce a lot of risk and stall improvements until then. Most importantly, it means that all future changes (and I am doing some right now) have to keep doing it the old, wrong way, making it harder, more effort and much riskier to eventually do the big rewriting.
In a 2.0, we can have a ServiceFactory instead of services, which is not _that_ big of a change and start doing it better right away.

### DIFF
--- a/src/main/java/de/retest/recheck/RecheckAdapter.java
+++ b/src/main/java/de/retest/recheck/RecheckAdapter.java
@@ -13,6 +13,15 @@ import de.retest.recheck.ui.descriptors.RootElement;
 public interface RecheckAdapter {
 
 	/**
+	 * Returns an instance of the RecheckAdapter, that has been initialized with the given {@link RecheckOptions}. In
+	 * case more specific recheck options are used during initialization of the {@link RecheckImpl}, then these are
+	 * passed on (and need to be cast).
+	 */
+	default RecheckAdapter initialize( final RecheckOptions opts ) {
+		return this;
+	}
+
+	/**
 	 * Returns {@code true} if the given object can be converted by the adapter.
 	 *
 	 * @param toVerify
--- a/src/main/java/de/retest/recheck/RecheckOptions.java
+++ b/src/main/java/de/retest/recheck/RecheckOptions.java
@@ -17,6 +17,7 @@ import de.retest.recheck.persistence.GradleProjectLayout;
 import de.retest.recheck.persistence.MavenProjectLayout;
 import de.retest.recheck.persistence.NamingStrategy;
 import de.retest.recheck.persistence.ProjectLayout;
+import de.retest.recheck.ui.descriptors.idproviders.RetestIdProvider;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.RequiredArgsConstructor;
@@ -32,6 +33,7 @@ public class RecheckOptions {
 	private final ProjectLayout projectLayout;
 	private final boolean reportUploadEnabled;
 	private final Filter filter;
+	private final RetestIdProvider retestIdProvider;
 
 	/**
 	 * Creates a shallow copy of the given options. Useful when extending the RecheckOptions to minimize dependencies on
@@ -58,7 +60,7 @@ public class RecheckOptions {
 	 */
 	protected RecheckOptions( final RecheckOptions toCopy ) {
 		this( toCopy.fileNamerStrategy, toCopy.namingStrategy, toCopy.projectLayout, toCopy.reportUploadEnabled,
-				toCopy.filter );
+				toCopy.filter, toCopy.retestIdProvider );
 	}
 
 	/**
@@ -88,6 +90,13 @@ public class RecheckOptions {
 	 */
 	public boolean isReportUploadEnabled() {
 		return reportUploadEnabled;
+	}
+
+	/**
+	 * The {@link RetestIdProvider} to use to generate the retestId for the elements in the Golden Masters.
+	 */
+	public RetestIdProvider getRetestIdProvider() {
+		return retestIdProvider;
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


roesslerj:
Btw., the RetestIdProvider is only used in downstream projects.

roesslerj:
See https://github.com/retest/recheck-web/compare/feature%2FomitScreenshots ...

src/main/java/de/retest/recheck/RecheckOptions.java
```diff
@@ -122,9 +132,11 @@ public NamingStrategy getNamingStrategy() {
 		private String suiteName = null;
 		private boolean reportUploadEnabled = false;
 		private Filter ignoreFilter = null;
+		private RetestIdProvider retestIdProvider = RetestIdProviderUtil.getConfiguredRetestIdProvider();
 		private final List<Filter> ignoreFilterToAdd = new ArrayList<>();
 
-		protected RecheckOptionsBuilder() {}
+		protected RecheckOptionsBuilder() {
```
diba1013: Your formatter seems off.

src/main/java/de/retest/recheck/RecheckAdapter.java
```diff
@@ -12,6 +12,15 @@
  */
 public interface RecheckAdapter {
 
+	/**
+	 * Returns an instance of the RecheckAdapter, that has been initialized with the given {@link RecheckOptions}. In
+	 * case more specific recheck options are used during initialization of the {@link RecheckImpl}, then these are
+	 * passed on (and need to be cast).
+	 */
+	default RecheckAdapter initialize( final RecheckOptions opts ) {
```
diba1013, line 20: So, this is essentially a factory method (as you mentioned, we should have a proper factory class for this). 

I would rename this method and javadoc to underline, that we expect a **new** RecheckAdapter (as setting fields is not the ideal way). I would even suggest to have this incorporated into `convert` as parameter (via another default method)?

```
default Set<RootElement> convert(Object toVerify, RecheckOptions options) {
    return check(toVerify);
}
```

Secondly, I am not sure, if we should expose all options to the adapter, as this may confuse people.

src/main/java/de/retest/recheck/execution/RecheckAdapters.java
```diff
@@ -4,18 +4,20 @@
...
 	private RecheckAdapters() {}
 
-	public static RecheckAdapter findAdapterFor( final Object toVerify ) {
+	public static RecheckAdapter findAdapterFor( final Object toVerify, final RecheckOptions options ) {
 		return StreamSupport.stream( adapters.spliterator(), false ) //
 				.filter( adapter -> adapter.canCheck( toVerify ) ) //
 				.findAny() //
-				.orElseThrow( () -> createHelpfulExceptionForMissingAdapter( toVerify.getClass().getName() ) );
+				.orElseThrow( () -> createHelpfulExceptionForMissingAdapter( toVerify.getClass().getName() ) )
+				.initialize( options );
```
diba1013, line 20: Be advised, that a `ServiceLoader` caches the classes. So you would call `initialize` multiple times&mdash;with its name suggest only be called once.

diba1013 (changes requested):

src/main/java/de/retest/recheck/RecheckAdapter.java
roesslerj: No, we do not want this in the convert method.
How would you rename the method?

src/main/java/de/retest/recheck/execution/RecheckAdapters.java
roesslerj: Yes, but I don't see that as a problem. Do you?

src/main/java/de/retest/recheck/RecheckAdapter.java
diba1013: > No, we do not want this in the convert method.

Now we create (ideally) a new instance of a `RecheckAdapter` for each `check`&mdash;though the options do not change with the lifetime of the `Recheck` instance&mdash;,which seems to be a bit excessive. Why not just pass this along with the check and have only one single adapter instance?

> How would you rename the method?

Well, now the initial `RecheckAdapter` which is created with the default constructor through the `ServiceLoader` only acts as a factory (as described above). Thus keep to some default like `create`.

src/main/java/de/retest/recheck/execution/RecheckAdapters.java
diba1013: I just wanted to emphasize that we should discuss the lifetime of a `RecheckAdapter` as with this proposal, it only lives during a single `check`.

roesslerj: I see you point, but for the Adapters we currently have, this is no problem, as they are lightweight. Anyways, the situation wouldn't be any different in a big refactoring, if this was a clean factory. It shouldn't be a problem now.

src/main/java/de/retest/recheck/RecheckAdapter.java
roesslerj: Hmm. I was just renaming this method when I realized that the current approach is better. We currently do not enforce creating a new instance. So if for some reason an adapter is heavyweight (which currently is not the case), it is up to the adapter to decide whether he wants to return a fresh instance here, or just change some internal values. I thus would keep the old method name. That also addresses your concern about the lifecycle.

diba1013: Could you comment on your aversion for an additional parameter. This would not break the API, save on objects being created and remove the need for a factory.

> So if for some reason an adapter is heavyweight (which currently is not the case)

We just do some IO and read the default.yml and attribtute.yml every time the adapter is created. I would not call this lightweight.

diba1013 (approved):

PR #466, opened 2019-10-15
Fix NPE as reported ...

roesslerj:
java.lang.NullPointerException
    at de.retest.recheck.report.ActionReplayResult.getCheckedUiElementsCount(ActionReplayResult.java:249)
    at de.retest.recheck.report.TestReplayResult.getCheckedUiElementsCount(TestReplayResult.java:84)
    at de.retest.recheck.report.SuiteReplayResult.getCheckedUiElementsCount(SuiteReplayResult.java:149)
    at de.retest.recheck.report.TestReport.getCheckedUiElementsCount(TestReport.java:121)
...

### DIFF
--- a/src/main/java/de/retest/recheck/report/ActionReplayResult.java
+++ b/src/main/java/de/retest/recheck/report/ActionReplayResult.java
@@ -246,7 +246,7 @@ public class ActionReplayResult implements Serializable {
 	}
 
 	public int getCheckedUiElementsCount() {
-		if ( !windows.isEmpty() ) {
+		if ( hasWindows() ) {
 			int result = 0;
 			for ( final RootElement rootElement : windows ) {
 				result += rootElement.countAllContainedElements();


modulo11:
Can you add a test for that?

modulo11 (dismissed):
Can you add a test for that?

src/main/java/de/retest/recheck/report/ActionReplayResult.java
```diff
@@ -246,7 +246,7 @@ public Screenshot getTargetScreenshot() {
 	}
 
 	public int getCheckedUiElementsCount() {
-		if ( !windows.isEmpty() ) {
+		if ( windows != null && !windows.isEmpty() ) {
```
diba1013: How can this happen anyways? The only way is through deserialization with `ActionReplayResult()` as the other constructor should prevent passing nulls.

Also, down below there is a method `hasWindows()`, if you like to use it.

diba1013 (approved):

PR #471, opened 2019-10-18
Feature/matcher consider parents

### LINKED ISSUE #465: Matcher doesn't ignore child elements if other than XPath
According to [our docs](https://docs.retest.de/recheck/usage/filter/#matching-elements):

>Elements are identified by one special attribute `$key`, so called _identifying attributes_. If you directly match an element with no additional expressions (e.g. attributes), all of its child elements are ignored too.
>
>```
># Match the element where its attribute $key fully matches the $value
>matcher: $key=$value
>```

This works fine when using an XPath, for example:

```
matcher: xpath=html[1]/[body[1]/div[1]
```

But when using a different matcher (i.e. `retestid`, `id`, `class` or `type`), child elements aren't ignored anymore. To verify this issue:

1. Pick an arbitrary page. 
1. Create a GM.
1. Change something (test fails).
1. Ignore the parent of the changed element via XPath.
1. Run again (test passes).
1. Change the ignore to something else as mentioned above.
1. Run again (test fails).

I assume this is because in `ElementXPathMatcher` we do:

```java
@Override
public boolean test( final Element element ) {
	return element.getIdentifyingAttributes().getPath().startsWith( xpath );
}
```

This includes child elements because they share the same XPath prefix. However, this isn't the case for `retestid`, `id`, `class` or `type`. Here, we don't check if a parent has been ignored.

A possible solution: Let all matchers also be a `ElementXPathMatcher` (e.g. via inheritance). Whenever such a matcher actually matches—for instance, `ElementIdMatcher` finds an element with an ID that should be ignored—then we store the XPath too and incorporate this in subsequent invocations. Note that:

1. A matcher might matches multiple elements, consequently, we would have to store and compare multiple XPaths.
2. This approach requires ordering. Assume we want to ignore `id=foo`, where this element's XPath equals `html[1]/[body[1]/div[1]/div[1]`. I don't believe this is the case, but if e.g. we check one of the child elements before we store the element-to-be-ignored's XPath, it won't be ignored.

Other ideas?

Creepow:
Let's talk about this on monday and find/agree on a maybe better solution.

Resolves #465

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Table of Contents
 
 * Do not parse the CSS attribute `box-shadow` for the filter `pixel-diff`.
 * Update screenshots always instead of never.
+* Matching elements now properly ignores child elements for identifying attributes other than XPath.
 
 ### New Features
 
--- a/src/main/java/de/retest/recheck/review/ignore/ElementFilter.java
+++ b/src/main/java/de/retest/recheck/review/ignore/ElementFilter.java
@@ -20,7 +20,12 @@ public class ElementFilter implements Filter {
 
 	@Override
 	public boolean matches( final Element element ) {
-		return matcher.test( element );
+		if ( matcher.test( element ) ) {
+			return true;
+		}
+		final Element parent = element.getParent();
+		return parent != null && matches( parent );
+
 	}
 
 	@Override
--- a/src/test/java/de/retest/recheck/review/ignore/matcher/ElementIdMatcherTest.java
+++ b/src/test/java/de/retest/recheck/review/ignore/matcher/ElementIdMatcherTest.java
@@ -1,0 +1,45 @@
+package de.retest.recheck.review.ignore.matcher;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import de.retest.recheck.ui.descriptors.Element;
+import de.retest.recheck.ui.descriptors.IdentifyingAttributes;
+
+class ElementIdMatcherTest {
+
+	ElementIdMatcher cut;
+
+	@BeforeEach
+	void setUp() {
+		final Element element = mock( Element.class );
+		final IdentifyingAttributes attribs = mock( IdentifyingAttributes.class );
+		when( element.getIdentifyingAttributes() ).thenReturn( attribs );
+		when( attribs.get( "id" ) ).thenReturn( "someDiv" );
+		cut = new ElementIdMatcher( element );
+	}
+
+	@Test
+	void should_match_when_id_is_equal() {
+		final Element element = mock( Element.class );
+		final IdentifyingAttributes attribs = mock( IdentifyingAttributes.class );
+		when( element.getIdentifyingAttributes() ).thenReturn( attribs );
+		when( attribs.get( "id" ) ).thenReturn( "someDiv" );
+
+		assertThat( cut ).accepts( element );
+	}
+
+	@Test
+	void should_not_match_when_id_is_not_equal() {
+		final Element element = mock( Element.class );
+		final IdentifyingAttributes attribs = mock( IdentifyingAttributes.class );
+		when( element.getIdentifyingAttributes() ).thenReturn( attribs );
+		when( attribs.get( "id" ) ).thenReturn( "otherDiv" );
+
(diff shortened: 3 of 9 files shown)


src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
```diff
@@ -25,12 +25,20 @@ private ElementClassMatcher( final List<String> classValues ) {
 	}
 
 	@Override
-	public boolean test( final Element element ) {
-		final List<String> classValues = toClassValuesList( toClassValueString( element ) );
-		if ( classValues.isEmpty() ) {
+	public boolean test( Element element ) {
+		final List<String> elementClassValues = toClassValuesList( toClassValueString( element ) );
```
beatngu13: Less verbose when using recursion:

```java
@Override
public boolean test( Element element ) {
    // actual code to test predicate
    final Element parent = element.getParent();
    return parent != null && test( parent );
}
```

And if we want to extract testing parents:

```java
public static boolean testParents( final Element element, final Matcher<Element> matcher ) {
    final Element parent = element.getParent();
    return parent != null && matcher.test( parent );
}
```

Which then can be called from each matcher like:

```java
public boolean test( Element element ) {
    // test predicate
    testParents( element, this );
}
```

src/test/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcherTest.java
```diff
@@ -13,62 +13,86 @@
 class ElementClassMatcherTest {
 
 	Element element;
-	ElementClassMatcher cut;
+	ElementClassMatcher matcher;
```
beatngu13: We always use `cut` to refer to the _class under test_ (see wiki), could you please revert this?

src/test/java/de/retest/recheck/review/ignore/matcher/ElementIdMatcherTest.java
```diff
@@ -0,0 +1,83 @@
...
+		when( element.getIdentifyingAttributes() ).thenReturn( attribs );
+		when( attribs.get( "id" ) ).thenReturn( "someDiv" );
+
+		when( childElement.getParent() ).thenReturn( element );
+
+		assertThat( matcher ).accepts( childElement );
+	}
+
+	@Test
+	void should_match_when_id_is_parent_of_parent() {
```
beatngu13: Do we need both tests? Parent and parent of parent?

beatngu13 (changes requested):

beatngu13:
Two additional, general remarks:

1. Does [this](https://docs.retest.de/recheck/usage/filter/#by-value) still work?
1. If you open a PR that corresponds to an issue, you can use a [closing keyword](https://help.github.com/en/articles/closing-issues-using-keywords) to a) link to the issue from the PR and vice versa, and b) automatically close the issue if the PR gets merged.

src/test/java/de/retest/recheck/review/ignore/matcher/ElementIdMatcherTest.java
Creepow: I would say yes. At one point i forgot the line `element = element.getParent();` which would lead to this test failing. I think we need the it to test all cases.

beatngu13 (changes requested):

beatngu13:
Oh, and could you please rebase and mark it "Ready for review"?

beatngu13:
I'm good with this. @diba1013?

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ Table of Contents
 
 * Do not parse the CSS attribute `box-shadow` for the filter `pixel-diff`.
 * Update screenshots always instead of never.
+* Matcher now does ignore child elements for not only XPath.
```
diba1013:
```suggestion
* Matching elements now properly ignores child elements for identifying attributes other than XPath.
```

src/main/java/de/retest/recheck/review/ignore/ElementFilter.java
```diff
@@ -20,7 +20,12 @@ public ElementFilter( final Matcher<Element> matcher ) {
 
 	@Override
 	public boolean matches( final Element element ) {
-		return matcher.test( element );
+		if ( matcher.test( element ) ) {
+			return true;
+		} else {
```
diba1013: You may omit the unnecessary `else`.

diba1013 (changes requested):

diba1013 (approved):

beatngu13 (approved):

PR #473, opened 2019-10-21
Let parent of `RootElement` be null

beatngu13:
The previous "Zombie" element caused several issues, because it was basically in an illegal state. I believe having a null parent in this case is the correct, or at least expected behavior (see e.g. #471). This is also what the comment from the empty `Element` ctor suggests: "Warning: Only to be used by JAXB!"

Does anyone know if this may cause issues elsewhere?


diba1013 (commented):
The code looks good, if it does not break anything. If @roesslerj has no objections, you may approve.

roesslerj (approved):

PR #474, opened 2019-10-21
Feature/sonarcloud changes

Creepow:


src/main/java/de/retest/recheck/RecheckOptions.java
```diff
@@ -242,10 +242,10 @@ public RecheckOptionsBuilder addIgnore( final String filterName ) {
 		}
 
 		public RecheckOptions build() {
-			final String suiteName = getSuiteName();
-			final NamingStrategy namingStrategy = new FixedSuiteNamingStrategy( suiteName, this.namingStrategy );
-			return new RecheckOptions( fileNamerStrategy, namingStrategy, projectLayout, reportUploadEnabled,
-					buildFilter( suiteName ) );
+			final String actualSuiteName = getSuiteName();
```
beatngu13: Everything looks really good, beside this here. I think this is something we can ignore in Sonar as I (personally) find it more confusing this way. Would you agree and, if you do, revert the commit?

beatngu13 (changes requested):

src/main/java/de/retest/recheck/RecheckOptions.java
diba1013: This shadowing was done somewhat intentionally, so that no one accidentally uses `this.suiteName`.

beatngu13 (approved):

PR #480, opened 2019-10-24
Otherwise we get an UIException if the file exists

roesslerj:


src/test/java/de/retest/recheck/util/ApprovalsUtil.java
```diff
@@ -153,6 +153,7 @@ public void increment() {}
 		@Override
 		public void report( final String received, final String approved ) {
 			try {
+				Paths.get( approved ).toFile().delete();
 				Files.copy( Paths.get( received ), Paths.get( approved ) );
```
beatngu13: The `copy` method allows you to pass `CopyOptions` like:

```java
Files.copy( Paths.get( received ), Paths.get( approved ), StandardCopyOption.REPLACE_EXISTING );
```

Then you can remove the inserted code.

beatngu13 (changes requested):

beatngu13 (approved):

PR #481, opened 2019-10-24
Feature/warning improvement

roesslerj:
This minor changes should be in the model of 1.6.0 for downstream implementation of functionality.

### DIFF
--- a/src/main/java/de/retest/recheck/ui/diff/ElementIdentificationWarning.java
+++ b/src/main/java/de/retest/recheck/ui/diff/ElementIdentificationWarning.java
@@ -19,14 +19,22 @@ public class ElementIdentificationWarning implements Serializable {
 	private static final long serialVersionUID = 1L;
 
 	@XmlAttribute
-	private final String testClassName;
+	private final String testFileName;
 
 	@XmlAttribute
 	private final Integer testLineNumber;
 
+	@XmlAttribute
+	private final String findByMethodName;
+
+	@XmlAttribute
+	private final String qualifiedTestName;
+
 	@SuppressWarnings( "unused" )
 	private ElementIdentificationWarning() {
-		testClassName = null;
+		testFileName = null;
 		testLineNumber = null;
+		findByMethodName = null;
+		qualifiedTestName = null;
 	}
 }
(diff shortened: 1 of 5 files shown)


diba1013 (commented):
As I am not up to date of the functionality, could you add some explanation (preferably in JavaDoc, as it is supposed to be public API?) on what the fields are supposed to do.

Also note, this is the second time, we refactor this warning stuff. Do we have any solid plan for the API and the actual implementation or is it, as always, just a blurry vision? While the general idea is known, I do not see the concrete plan or beginnings, on how recheck is supposed to realize this. _Therefore, I would prefer a workshop where you explain your thoughts, let us design a proper API and afterwards realize this change._

roesslerj:
You are right ... you are not in the depths of the topic. Sorry I assigned the PR to you. Should have referred to Daniel instead. Will correct.
To clarify: The first refactoring was based on Ivans Implementation. As I adjusted that, I saw that we need to have another go. So this is the second iteration. Overall, the approach is called _agile_. And I find your lack of faith disturbing ;-)

src/main/java/de/retest/recheck/ui/diff/ElementIdentificationWarning.java
```diff
@@ -19,14 +19,22 @@
...
+	private final String testFileName;
 
 	@XmlAttribute
 	private final Integer testLineNumber;
 
+	@XmlAttribute
+	private final String findByMethodName;
+
+	@XmlAttribute
+	private final String qualifiedTestName;
```
beatngu13, line 50: What is the difference between `testFileName` and this? For what do we need which information?

roesslerj: When implementing in the downstream projects, I realized that there can be test classes that life in files of which the names differ. So you could (in theory) have a TestXY that lives in a de.retest.Tests class. Also, you could have multiple Tests.java files, that differ in package/path. Will document via JavaDoc.

beatngu13: So, `testFileName` is just the name of the source file (e.g. `MyTest` or `MyIT`), whereas `qualifiedTestName` is the name of the fully-qualified test class name (e.g. `com.example.MyTest` or `com.example.MyTest.MyNestedTest`)?

beatngu13: Ah, I see. 👍 As you said, some minimal Javadoc would be helpful, maybe with an actual example.

beatngu13 (approved):

PR #484, opened 2019-10-25
Add a convenience method to return the value of an attribute

roesslerj:
no matter whether it's an identifying or "regular" attribute. I've stumbled across similar code a couple of times, and this simplifies it.

### DIFF
--- a/src/main/java/de/retest/recheck/ui/descriptors/Element.java
+++ b/src/main/java/de/retest/recheck/ui/descriptors/Element.java
@@ -219,13 +219,17 @@ public class Element implements Serializable, Comparable<Element> {
 	 * @return The {@link java.io.Serializable} value of the attribute.
 	 */
 	public Object getAttributeValue( final String attributeName ) {
-		if ( getIdentifyingAttributes().get( attributeName ) != null ) {
-			return getIdentifyingAttributes().getAttribute( attributeName ).getValue();
-		} else if ( getAttributes().get( attributeName ) != null ) {
-			return getAttributes().get( attributeName );
-		} else {
-			return null;
+		final Attribute identifyingAttribute = getIdentifyingAttributes().getAttribute( attributeName );
+		if ( identifyingAttribute != null ) {
+			return identifyingAttribute.getValue();
 		}
+
+		final Object ordinaryAttribute = getAttributes().get( attributeName );
+		if ( ordinaryAttribute != null ) {
+			return ordinaryAttribute;
+		}
+
+		return null;
 	}
 
 	public String getRetestId() {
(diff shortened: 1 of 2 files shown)


src/main/java/de/retest/recheck/ui/descriptors/Element.java
```diff
@@ -213,6 +213,21 @@ public Attributes getAttributes() {
 		return attributes;
 	}
 
+	/**
+	 * Returns the value of the given attribute, whether this is an identifying attribute or not.
+	 *
+	 * @return The {@link java.io.Serializable} value of the attribute.
+	 */
+	public Object getAttributeValue( final String attributeName ) {
```
beatngu13, line 221: When it's `Serializable` as stated in the Javadoc, does it make sense to adapt the return type accordingly?

beatngu13, line 221: The implementation "prefers" identifying over ordinary attributes. It might be that we have the same key/name in both sets, for example `type` in recheck-web. Here, the identifying attribute is the tag name, whereas the ordinary attribute refers to the corresponding [HTML attribute](https://www.w3schools.com/tags/att_type.asp).

While the implementation is fine, I would suggest to add a test for this behavior (e.g. `getAttributeValue_should_prefer_identifying_over_ordinary_attribute`) in case someone accidentially changes this.

beatngu13, line 221: (Optional) suggestion to avoid repeated calls to getters:

```java
public Object getAttributeValue( final String attributeName ) {
	Object identifyingAttribute = getIdentifyingAttributes().get( attributeName );
	if ( identifyingAttribute != null ) {
		return identifyingAttribute.getValue();
	}

	Object ordinaryAttribute = getAttributes().get( attributeName );
	if ( ordinaryAttribute != null ) {
		return ordinaryAttribute.getValue(); // Doesn't this require getValue() too?
	}

	return null;
}
```

beatngu13 (changes requested):

beatngu13:
Good idea, I did this multiple times too.

diba1013:
This is a good idea as we want to unify them anyways.

src/main/java/de/retest/recheck/ui/descriptors/Element.java
diba1013: Or even generic `T extends Serializable` or simply `T` to avoid casting at ten different places.

roesslerj: Done

roesslerj: There is no common return type `Serializable`. Adapting that would be out of scope of this PR. I do not understand how I could return a generic here...

beatngu13 (approved):

PR #487, opened 2019-10-28
Feature/handle no gm exception

### LINKED ISSUE #145: Commit fails when a single GM cannot be found
When applying changes via:

```
$ recheck commit --all my.report
```

The command fails when a single GM cannot be found.

Suppose there are 100 GMs affected from changes and a single GM has been deleted. If the changes for this GM appear at the beginning of the report, all remaining GMs aren't updated with a message similar like this:

```
Test report 'ad405f44-9421-4dbf-b4df-df13b5015061.report' has 1936 difference(s) in 27 test(s).
Updated Golden Master '/Users/kruse/dev/workspaces/private/visual-testing/src/test/resources/retest/recheck/com.github.beatngu13.visualtesting.RecheckWebTest/chrome-360.cn.initial.recheck'.
Updated Golden Master '/Users/kruse/dev/workspaces/private/visual-testing/src/test/resources/retest/recheck/com.github.beatngu13.visualtesting.RecheckWebTest/chrome-apple.com.initial.recheck'.
Updated Golden Master '/Users/kruse/dev/workspaces/private/visual-testing/src/test/resources/retest/recheck/com.github.beatngu13.visualtesting.RecheckWebTest/chrome-bbc.com.initial.recheck'.
The Golden Master 'src/test/resources/retest/recheck/com.github.beatngu13.visualtesting.RecheckWebTest/chrome-taobao.com.initial.recheck' cannot be found.
Please make sure that the given test report '/Users/kruse/dev/workspaces/private/visual-testing/ad405f44-9421-4dbf-b4df-df13b5015061.report' is within the corresponding project directory.
```

I think we should try to apply as many updates as possible and not immediately stop. Or at least guarantee [atomicity](https://en.wikipedia.org/wiki/ACID#Atomicity). Or even better: Ask the user what he wants to do like:

```
The following Golden Masters cannot be found:
    ...
    ...
    ...
Do you want to apply the remaining changes (yes/y) or abort (no/n)?
```

`Commit` uses `ApplyChangesToStatesFlow` from recheck, which throws a `NoGoldenMasterFoundException` in this case, so this change probably affects recheck too.

Creepow:
Resolves https://github.com/retest/recheck.cli/issues/145

### DIFF
--- a/src/main/java/de/retest/recheck/persistence/NoGoldenMasterFoundException.java
+++ b/src/main/java/de/retest/recheck/persistence/NoGoldenMasterFoundException.java
@@ -4,6 +4,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class NoGoldenMasterFoundException extends Exception {
 
@@ -12,9 +13,12 @@ public class NoGoldenMasterFoundException extends Exception {
 	private final List<String> filenames;
 
 	public NoGoldenMasterFoundException( final String... filenames ) {
-		super( "No Golden Master file(s) with the following name(s) found: \n"
-				+ Arrays.asList( filenames ).stream().collect( Collectors.joining( "\n" ) ) );
+		super( "No Golden Master file(s) with the following name(s) found:\n" + listFilenames( filenames ) );
 		this.filenames = Arrays.asList( filenames );
+	}
+
+	private static String listFilenames( final String[] filenames ) {
+		return Stream.of( filenames ).collect( Collectors.joining( "\n\t", "\t", "" ) );
 	}
 
 	public List<String> getFilenames() {
--- a/src/main/java/de/retest/recheck/persistence/NoGoldenMasterFoundException.java
+++ b/src/main/java/de/retest/recheck/persistence/NoGoldenMasterFoundException.java
@@ -3,6 +3,7 @@ package de.retest.recheck.persistence;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class NoGoldenMasterFoundException extends Exception {
 
@@ -11,7 +12,8 @@ public class NoGoldenMasterFoundException extends Exception {
 	private final List<String> filenames;
 
 	public NoGoldenMasterFoundException( final String... filenames ) {
-		super( "No Golden Master with the following name(s) has/have been found: " + Arrays.toString( filenames ) );
+		super( "No Golden Master file(s) with the following name(s) found: \n"
+				+ Arrays.asList( filenames ).stream().collect( Collectors.joining( "\n" ) ) );
 		this.filenames = Arrays.asList( filenames );
 	}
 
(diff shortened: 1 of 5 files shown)


src/main/java/de/retest/recheck/persistence/NoGoldenMasterFoundException.java
```diff
@@ -1,17 +1,21 @@
...
 	private static final long serialVersionUID = 1L;
 
-	private final String filename;
+	private final List<String> filenames;
 
-	public NoGoldenMasterFoundException( final String filename ) {
-		super( "No Golden Master with the name '" + filename + "' has been found!" );
-		this.filename = filename;
+	public NoGoldenMasterFoundException( final String... filenames ) {
+		super( "No Golden Master with the following name(s) has/have been found: " + Arrays.toString( filenames ) );
```
beatngu13: The `Arrays#toString(...)` variant was just to make it work. As described in https://github.com/retest/recheck.cli/issues/145, I thought about something like:

```
No Golden Master file(s) with the following name(s) found:
    foo
    bar
    baz
```

Have a look at [`Collectors#joining(...)`](https://www.concretepage.com/java/jdk-8/java-8-collectors-joining-example), this might be helpful here.

src/main/java/de/retest/recheck/suite/flow/ApplyChangesToStatesFlow.java
```diff
@@ -22,6 +22,7 @@
 	private static final Logger logger = LoggerFactory.getLogger( ApplyChangesToStatesFlow.class );
 
 	private final GoldenMasterProvider goldenMasterProvider;
+	private final List<String> filenames = new ArrayList<>();
```
beatngu13: `filenames` of what? 😉 Without context, this is not self-explanatory. I suggest something like `missingGoldenMastersFilenames`.

src/test/java/de/retest/suite/flow/ApplyChangesToStatesFlowTest.java
```diff
@@ -78,4 +80,66 @@ public void apply_should_apply_changes_to_states() throws Exception {
...
+
+		final SutState after2 = persistence.load( temporaryFile2.toURI() );
+		final Element changed2 = after2.getRootElements().get( 0 ).getContainedElements().get( 0 )
+				.getContainedElements().get( 0 ).getContainedElements().get( 0 ).getContainedElements().get( 0 )
+				.getContainedElements().get( 0 );
+
+		assertThat( changed2.getAttributes().get( "enabled" ) ).isEqualTo( true );
+	}
+
+	private SuiteChangeSet buildChanges( final File temporaryFile1, final File temporaryFile2,
```
beatngu13: Now you only use this method in the new test case, right? Instead of doing the same thing three times, have a single method and reuse it. Ideally, you also reuse it in the other test case. Remember [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

beatngu13 (changes requested):
There is also a failing test case:

https://travis-ci.com/retest/recheck/builds/133850028#L3198

Here you only have to adapt the expected string.

beatngu13 (approved):

PR #492, opened 2019-10-31
Fix/properly format no golden master

diba1013:


src/main/java/de/retest/recheck/persistence/NoGoldenMasterFoundException.java
```diff
@@ -6,19 +6,26 @@
...
 	public NoGoldenMasterFoundException( final String... filenames ) {
-		super( "No Golden Master file(s) with the following name(s) found:\n" + listFilenames( filenames ) );
+		super( "The following Golden Master(s) cannot be found:" + listFilenames( filenames ) );
 		this.filenames = Arrays.asList( filenames );
 	}
 
 	private static String listFilenames( final String[] filenames ) {
-		return Stream.of( filenames ).collect( Collectors.joining( "\n\t", "\t", "" ) );
+		if ( ArrayUtils.isEmpty( filenames ) ) {
+			throw new IllegalArgumentException( "You should at least provide one golden master file." );
```
beatngu13:
```suggestion
			throw new IllegalArgumentException( "You should at least provide one Golden Master file." );
```

src/test/java/de/retest/suite/flow/CreateChangesetForAllDiffrencesFlowTest.java
```diff
@@ -98,11 +98,11 @@ public void apply_should_throw_exception_whitout_the_stateFilePath() throws Exce
 
 		assertThatThrownBy( () -> ApplyChangesToStatesFlow.apply( persistence, changesWhitoutStateFilePath ) ) //
 				.isInstanceOf( NoGoldenMasterFoundException.class ) //
-				.hasMessage( "No Golden Master file(s) with the following name(s) found:\n\tnull" );
+				.hasMessage( "The following Golden Master(s) cannot be found:\n\t- null" );
```
beatngu13, line 101: Shouldn't `ArrayUtils` prevent this from being null?

beatngu13 (changes requested):

src/test/java/de/retest/suite/flow/CreateChangesetForAllDiffrencesFlowTest.java
diba1013: There is a difference when passing nothing (i.e. `null`) or the Array `[ null ]`

beatngu13 (approved):

PR #494, opened 2019-11-01
Add some attributes that should be ignored by default...

roesslerj:
Please discuss and if you agree, include in release.
Anything else to add?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ Table of Contents
 
 * `RecheckImpl` can now be initialized as a member variable without additional parameters if inside a test class.
 * The screenshot does now have a fixed filename. This makes it much easier to work with VCS.
+* Add sensible default values to the recheck.ignore file that is installed. This will only be created with new project.
 
 
 --------------------------------------------------------------------------------
--- a/src/test/java/DefaultRecheckIgnoreTest.java
+++ b/src/test/java/DefaultRecheckIgnoreTest.java
@@ -1,0 +1,25 @@
+import static de.retest.recheck.ignore.FilterLoader.load;
+import static java.nio.file.Paths.get;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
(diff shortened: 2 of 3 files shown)


diba1013 (commented):
First of all, changelog (i.e. DoD). I assume, we have a test that loads this file and checks if there are invalid filter rules defined?

src/test/java/DefaultRecheckIgnoreTest.java
```diff
@@ -0,0 +1,30 @@
...
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.read.ListAppender;
+import de.retest.recheck.review.ignore.io.ErrorHandlingLoader;
+
+class DefaultRecheckIgnoreTest {
+
+	@Test
+	void default_recheck_ignore_should_load_without_exception() throws Exception {
+		final ListAppender<ILoggingEvent> listAppender = new ListAppender<>();
```
diba1013: We have an [extension](https://github.com/retest/recheck/blob/bd0f8ded6cb6cb4b98dcc466e5b622170a4db660/src/test/java/de/retest/recheck/review/workers/LoadAndSaveFilterWorkerTest.java#L29) that does this.

diba1013 (changes requested):

src/test/java/DefaultRecheckIgnoreTest.java
roesslerj: Wow, didn't know! Thanks

CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ Table of Contents
 
 ### Improvements
 
+* Add sensible default values to the recheck.ignore file that is installed.
```
diba1013: Shouldn't this be in the 1.6.0 section?

diba1013 (changes requested):

diba1013 (approved):
I did some additional changes and approved so that this in the next release.

PR #497, opened 2019-11-02
Feature/improve internal structure

roesslerj:
For usage in microservices etc.

### DIFF
--- a/src/main/java/de/retest/recheck/RecheckImpl.java
+++ b/src/main/java/de/retest/recheck/RecheckImpl.java
@@ -130,8 +130,7 @@ public class RecheckImpl implements Recheck, SutStateLoader {
 
 	protected ActionReplayResult createActionReplayResult( final Object toVerify, final RecheckAdapter adapter,
 			final String currentStep ) {
-		final DefaultValueFinder defaultFinder = adapter.getDefaultValueFinder();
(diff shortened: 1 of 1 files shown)


diba1013 (commented):
Could you please provide some context, why these changes are necessary. We now expose (i.e. protected) so much stuff, that it makes more sense to re-implement this within your micro-service, as the important classes should be in separate classes (e.g. `RecheckSutState` or `RecheckTestReportUtil`).

As stated, extending to change the behavior of a class is bad written code. I will approve, but it should be noted, that in the future, we should make this stuff configurable.

roesslerj:
Well, if I have to copy a code three times, it makes sense to remove the redundancy.

diba1013:
> Well, if I have to copy a code three times, it makes sense to remove the redundancy.

That's a classic violation of the [single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle) and denotes that our `RecheckImpl` does too much (by itself).

However, **I will gladly approve**, if you could give some more context (i.e. point me to a branch or commit in our services) so that I am able to see these changes. Since essential all methods are now either public or protected which seems like bad written code for an implementation of a interface.

As mentioned above, this was just a comment in that terms that we need to keep an eye on that and improve it once we got time. **This was no request for change!**

diba1013 (approved):

PR #505, opened 2019-11-06
Feature/js filter log cache

roesslerj:


diba1013 (changes requested):
Please share your thoughts to my [comment](https://github.com/retest/recheck/pull/498#issuecomment-549088081), since especially now that we've got time.

diba1013 (approved):

PR #507, opened 2019-11-07
Add the code of conduct

diba1013:
This is the default template provided from GitHub that goes together with our company values.


diba1013:
A thought: To remove redundancy, we could also move this into the docs and keep a minimal file here, referencing the essential parts and refer to the docs page.

roesslerj:
Hmm. We really probably won't change it all that much. So I think the pain of the redundancy should be very low ... so we'll keep it—like the license.

roesslerj (approved):

PR #512, opened 2019-11-08
Feature/rework filter loader mechanism

diba1013:

- [x] ~you updated the changelog.~

## Description

> TL;DR This PR addresses several issues with the current implementation of the _Loading_ API with respects to type safety, API and error handling.

#### Type Safety

The `Loaders` implementation has automatic generics which get adjusted accordingly to the left hand of the expression (see below). Together with stuffing all possible loaders into one List this resulted in a "I have to guess the respective type" and if the guess was wrong, in a weird `ClassCastException`, since the last would always be a `PreserveLineFilter`.
```
String loaded = Loaders.load( "foo bar" );
```

To work around this issue, several independent lists of loaders per type have been created. Since those are essentially a loader for the common super type, a new Loader was created that does the delegation as it was used to by `Loaders`. The respective type specific Loaders are accessible in `Loaders`.

This change even allows for wrapped generics (e.g. `Matcher<Element>`) to be resolved properly.

#### API

The `Loader`-API featured two methods that had to be called when trying to load something. Firstly, `canLoad` which essentially did a dry run and if successful the actual `load` which does the same logic again.

These methods have been merged to return an `Optional` which directly contains the value if successful, or empty if unsuccessful. While this now has the advantages and disadvantages of `Optional`, it essentially provides a performance benefit. In particular, this is true for the `RegexLoader` which must only match once (updating the internal status) and instantly retrieve the result with `group`.

The stream loading has been moved into the `Loader` as default methods.

#### Error Handling

As noted above, determining when a filter can load the line is tedious. The same goes for error handling. What happens, if the loader wraps another one (like `ElementAttributeFilterLoader`) and it fails to load the child? The loader already claimed that it could load the line, but then fails to, which prevents other loaders to take effect.
```
matcher: foo=bar, attribute: bar
```

Due to the API changes, this is now possible. The child will try to load the respective subset (e.g. `foo=bar`) and be chained to the parent loader. If the child fails, the parent does to, which gives access to other loaders or the fallback, if all loaders fail.

### State of this PR

This should only be an internal refactoring, since `Loaders` should not be addressed from an external API. Instead the `Filters` and `*FilterWorker` should be used, which have been updated accordingly.

All necessary tests should have been updated and fixed accordingly. However with SonarCloud currently down, there is no certain way to tell.

### Additional Context

Please read the commit messages for additional information.

### DIFF
--- a/src/main/java/de/retest/recheck/review/ignore/io/Loader.java
+++ b/src/main/java/de/retest/recheck/review/ignore/io/Loader.java
@@ -1,10 +1,22 @@
 package de.retest.recheck.review.ignore.io;
 
 import java.util.Optional;
+import java.util.stream.Stream;
+
+import de.retest.recheck.util.StreamUtil;
 
 public interface Loader<T> {
 
 	Optional<T> load( String line );
 
+	default Stream<T> load( final Stream<String> lines ) {
+		return lines.map( this::load ) //
+				.flatMap( StreamUtil::optionalToStream );
+	}
+
 	String save( T ignore );
+
+	default Stream<String> save( final Stream<T> objects ) {
+		return objects.map( this::save );
+	}
 }
--- a/src/main/java/de/retest/recheck/review/ignore/io/Loader.java
+++ b/src/main/java/de/retest/recheck/review/ignore/io/Loader.java
@@ -1,11 +1,10 @@
 package de.retest.recheck.review.ignore.io;
 
+import java.util.Optional;
+
 public interface Loader<T> {
 
-	boolean canLoad( final String line );
-
-	T load( String line );
+	Optional<T> load( String line );
 
 	String save( T ignore );
-
 }
--- a/src/test/java/de/retest/recheck/review/ignore/io/InheritanceLoaderTest.java
+++ b/src/test/java/de/retest/recheck/review/ignore/io/InheritanceLoaderTest.java
@@ -1,0 +1,95 @@
+package de.retest.recheck.review.ignore.io;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class InheritanceLoaderTest {
+
+	Loader<A> cut;
+
+	final A a = new A();
+	final B b = new B();
+	final C c = new C();
+	final Z z = new Z();
+
+	@BeforeEach
+	void setUp() {
+		final Loader<A> aLoader = mock( Loader.class );
+		when( aLoader.load( "a" ) ).thenReturn( Optional.of( a ) );
+		when( aLoader.save( a ) ).thenReturn( "a" );
+
+		final Loader<B> bLoader = mock( Loader.class );
+		when( bLoader.load( "b" ) ).thenReturn( Optional.of( b ) );
+		when( bLoader.save( b ) ).thenReturn( "b" );
+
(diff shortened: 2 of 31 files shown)


src/test/java/de/retest/recheck/review/ignore/io/LoadersTest.java
```diff
@@ -0,0 +1,11 @@
+package de.retest.recheck.review.ignore.io;
+
+import org.junit.jupiter.api.Test;
+
+class LoadersTest {
+
+	@Test
+	void save_should_throw_exception_when_loader_cannot_be_found() throws Exception {
+
```
beatngu13: ?

src/test/java/de/retest/recheck/review/ignore/io/InheritanceLoaderTest.java
```diff
@@ -0,0 +1,94 @@
...
+	@Test
+	void save_with_unsupported_loader_should_produce_exception() throws Exception {
+		assertThatThrownBy( () -> cut.save( z ) ) //
+				.isInstanceOf( UnsupportedOperationException.class ) //
+				.hasMessage( String.format( "Did not find a loader for %s.", z ) );
+	}
+
+	@Test
+	void save_stream_with_unsupported_loader_should_produce_exception() throws Exception {
+		assertThatThrownBy( () -> cut.save( Stream.of( a, z, b ) ).count() ) // We need a finalizer here
```
beatngu13: Two things:

* Doesn't throw an exception: https://travis-ci.com/retest/recheck/builds/135640725#L3163
* Maybe convert your comment into an actual `TODO` and elaborate what you have in mind.

beatngu13 (changes requested):
Overall, great improvement! 👍

src/main/java/de/retest/recheck/review/ignore/io/Loader.java
```diff
@@ -1,11 +1,21 @@
 package de.retest.recheck.review.ignore.io;
 
+import java.util.Optional;
+import java.util.stream.Stream;
+
 public interface Loader<T> {
 
-	boolean canLoad( final String line );
```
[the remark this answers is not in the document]
diba1013: I should have removed all `canLoad` methods except from `FilterPreserveLineLoader` since this has a more complex `canLoad` method. However, I forgot to make it private, which I have now done.

src/test/java/de/retest/recheck/review/ignore/io/LoadersTest.java
diba1013: That was accidentally commited...

src/test/java/de/retest/recheck/review/ignore/io/InheritanceLoaderTest.java
diba1013: > Maybe convert your comment into an actual TODO

This was intentionally and is no TODO. It should have noted that the `save` only does a stream intermediate step; it does not evaluate the stream. Thus a finalizing method is needed, so that the stream is evaluated. I updated the comment to be hopefully more descriptive.

> Doesn't throw an exception

My mistake probably was that `count` might not evaluate the stream properly, if it can be retrieved by an other method (i.e. backing `list.size()`) (see [here](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Stream.html#count())). This seems to be an difference between my OracleJDK and the OpenJDK used in tests. I changed the finalizer to a `toArray` which must evaluate the Stream properly.

src/main/java/de/retest/recheck/review/ignore/io/Loader.java
```diff
@@ -1,11 +1,21 @@
...
+	Optional<T> load( String line );
 
-	T load( String line );
+	default Stream<T> load( final Stream<String> lines ) {
+		return lines.map( this::load ) //
```
beatngu13, line 13: Sorry, looks like my comment got lost or I've forgot to send … you can use our util here:

```java
return lines.map( this::load ).flatMap( StreamUtil:optionalToStream );
```

beatngu13 (changes requested):

beatngu13 (approved):

PR #515, opened 2019-11-13
Feature/also load ignore.js

roesslerj:
Add test and fix error


diba1013:
This should not happen! We are only looking for filters, not ignore files. Thus we only want to load `.filter.js` files and not all `.js` files.

If you worry/feel that we are not loading the default ignore, that is a error in `LoadFilterWorker`, which&mdash;despite its name&mdash;loads the `GlobalIgnoreApplier`.

PR #517, opened 2019-11-13
Make sure that the default-recheck.ignore.js file loaded without errors

roesslerj:
Just add a test that makes sure the default.ignore.js has no errors...

### DIFF
--- a/src/test/java/DefaultRecheckIgnoreJsTest.java
+++ b/src/test/java/DefaultRecheckIgnoreJsTest.java
@@ -1,0 +1,27 @@
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.nio.file.Paths;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.slf4j.event.Level;
+
+import de.retest.recheck.ignore.JSFilterImpl;
+import io.github.netmikey.logunit.api.LogCapturer;
+
+public class DefaultRecheckIgnoreJsTest {
+
+	@RegisterExtension
+	LogCapturer warningAndErrorLogs = LogCapturer.create() //
+			.forLevel( Level.WARN ) //
+			.captureForType( JSFilterImpl.class );
+
+	@Test
+	void default_recheck_ignore_should_load_without_exception() throws Exception {
+		new JSFilterImpl( Paths.get( getClass().getResource( "default-recheck.ignore.js" ).toURI() ) );
+
+		// Assert that no error was logged...
+		assertThat( warningAndErrorLogs.size() ).isEqualTo( 0 );
+	}
+
+}


src/test/java/DefaultRecheckIgnoreJsTest.java
```diff
@@ -0,0 +1,27 @@
...
+public class DefaultRecheckIgnoreJsTest {
+
+	@RegisterExtension
+	LogCapturer warningAndErrorLogs = LogCapturer.create() //
+			.forLevel( Level.WARN ) //
+			.captureForType( JSFilterImpl.class );
+
+	@Test
+	void default_recheck_ignore_should_load_without_exception() throws Exception {
+		new JSFilterImpl( Paths.get( getClass().getResource( "default-recheck.ignore.js" ).toURI() ) );
```
diba1013, line 21: Do we want to execute this filter or just load it.

src/test/java/DefaultRecheckIgnoreJsTest.java
```diff
@@ -0,0 +1,27 @@
...
+import org.slf4j.event.Level;
+
+import de.retest.recheck.ignore.JSFilterImpl;
+import io.github.netmikey.logunit.api.LogCapturer;
+
+public class DefaultRecheckIgnoreJsTest {
+
+	@RegisterExtension
+	LogCapturer warningAndErrorLogs = LogCapturer.create() //
+			.forLevel( Level.WARN ) //
```
diba1013, line 16: Should this also be `Level.ERROR` to catch any potential errors in [JSFilterImpl.java#L44](https://github.com/retest/recheck/blob/master/src/main/java/de/retest/recheck/ignore/JSFilterImpl.java#L44)?

diba1013 (changes requested):

src/test/java/DefaultRecheckIgnoreJsTest.java
roesslerj: Can we combine multiple levels? Because I tried it with a faulty file, and the test failed.

src/test/java/DefaultRecheckIgnoreJsTest.java
roesslerj: Apparently loading suffices to find erroneous files. As of now, filter only contains comments.

src/test/java/DefaultRecheckIgnoreJsTest.java
diba1013: I don't know the API in that regard. I would assume that ERROR would also catch WARN?

roesslerj: You mean the other way around. If you e.g. define a filter in the log config, you define the _lowest_ level of interest.

diba1013: Yes!? 😕

roesslerj: So I assume if we filter for WARN, everything _below_ will be filtered.

diba1013 (approved):

PR #520, opened 2019-11-18
Feature/add version check to persistence

diba1013:

- [X] the necessary tests are either created or updated.
- [X] all status checks (Travis CI, SonarCloud, etc.) pass.
- [X] you updated the changelog.
- [X] ~you updated necessary documentation within [retest/docs](https://github.com/retest/docs).~

## Description

> TL;DR This PR should provide the means to disable the loading of a report when breaking changes occur.

There are several use cases that can happen while loading the report:
1. The report can be loaded and is valid. We can freely work with the report and our model.
2. The report can be loaded, but is in an invalid state. This may happen when we technically do not alter the persisted values, but alter their meaning. This was done with #473 where the report still can be loaded, but the parent now is always null instead of containing an invalid zombie element. Thus the model already was trimmed to expect the `RootElement` to gave a `null`-parent.
3. The report is not loadable. This is due to changes in the persistence (i.e. zipping or altering the persistable values)

The second case was not handled properly before. The persistence did not encounter any exceptions while loading, thus assuming the report is valid. However due to said changes, selecting a filter due to  #471 resulted in a NPE.

We decided against parsing the already persisted `recheck-version` since this would require some more complex checks. Firstly getting the actual version (which could be transformed to an integer `"1.6.0" -> int 160` and then simply do the same check. But that would rise the following questions:
1. How do we specify which version not to load? Only load the current?
2. How do we treat snapshots, betas, hotfixes?
3. There is no version available through the IDE. How do we handle that?

This PR reuses the `persistenceVersion` like it was formerly used in XML to transform incompatible versions. However, since Kryo does not have such a transformation property, it is instead used to indicate the version compatible with the model. This then disallows lower versions to be returned, once successfully loaded.

### State of this PR

We have to keep an eye out, that every breaking change in the model should always increase the `persistenceVersion`. Should this also be the case for the `serialVersionUID`? Should we connect these both versions by `serialVersionUID  = PERSISTENCE_VERSION`?

This is currently only implemented for TestReports, since we generally assume that we only load reports with the `KryoPersistence`. However, it should be internally possible to add additional `Persistable` to the provided map.

`isUnknownFormat` still does some not comprehensible magic. That will be part of another PR. I even could have split this PR, but that would require merging and scheduling of the PR creation which would mean overhead for me.

### Additional Context

As this PR also acts as a bug fix (the last three commits), it disables the loading of all old reports due to a version increase of the `TestReport`. This acts as a bug fix for the reports generated by the chrome extension using an old recheck version and thus forcefully creates the second use case from above.


src/main/java/de/retest/recheck/ioerror/ReTestIOException.java
```diff
@@ -8,10 +8,20 @@
 	private static final long serialVersionUID = 1L;
 	private final URI location;
 
+	public ReTestIOException( final URI location ) {
```
beatngu13: If the URI-only ctors (also in subclasses) isn't effectively used, I wouldn't add it just for the sake of completeness.

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -35,6 +38,17 @@
 
 	private static final String OLD_RECHECK_VERSION = "an old recheck version (pre 1.5.0)";
 
+	private static final Map<Class<?>, Integer> versions = createVersions();
```
beatngu13: Maybe `compatibleVersions` or something like this? Just to make it easier to understand what these key-value pairs are for.

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -106,6 +126,15 @@ public T load( final URI identifier ) throws IOException {
 		}
 	}
 
+	private boolean isCompatible( final T persistable ) {
+		return isCompatible( persistable.getClass(), persistable.version() );
+	}
+
+	boolean isCompatible( final Class<? extends Persistable> clazz, final int version ) {
```
beatngu13, line 134: We still have no guideline for this, but maybe a comment that says it is for testing? (I think Guava also has an annotation for this, at some point we might want to roll out our own `@TestingOnly`.)

src/main/java/de/retest/recheck/report/TestReport.java
```diff
@@ -24,7 +24,7 @@
 public class TestReport extends Persistable {
 
 	private static final long serialVersionUID = 1L;
-	private static final int PERSISTENCE_VERSION = 20;
+	public static final int PERSISTENCE_VERSION = 20;
```
beatngu13: Isn't package private sufficient?

src/test/java/de/retest/recheck/persistence/bin/KryoPersistenceTest.java
```diff
@@ -92,4 +93,26 @@ void on_error_file_should_be_deleted() throws IOException {
 				.isInstanceOf( KryoException.class );
 		assertThat( nonexistent ).doesNotExist();
 	}
+
+	@Test
+	void isCompatible_should_be_compatible_with_equal_or_higher_value() throws Exception {
+		final KryoPersistence<TestReport> cut = new KryoPersistence<>();
```
beatngu13, line 110: Left-hand side can use the interface `Persistence`.

src/test/java/de/retest/recheck/persistence/bin/KryoPersistenceTest.java
```diff
@@ -92,4 +93,26 @@ void on_error_file_should_be_deleted() throws IOException {
 				.isInstanceOf( KryoException.class );
 		assertThat( nonexistent ).doesNotExist();
 	}
+
+	@Test
+	void isCompatible_should_be_compatible_with_equal_or_higher_value() throws Exception {
```
beatngu13, line 109: Maybe merge these tests into a parameterized one?

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ Table of Contents
 
 ### Breaking Changes
 
+* Reports before version 1.7.0 cannot be loaded anymore.
```
beatngu13: Maybe add: "Simply re-run your tests with the new recheck version to create them again."

beatngu13 (changes requested):

src/main/java/de/retest/recheck/report/TestReport.java
diba1013: I use that in `KryoPersistence`, so that is not possible.

src/test/java/de/retest/recheck/persistence/bin/KryoPersistenceTest.java
diba1013: I am using `isCompatible` below which requires the KryoPersistence. This is also in line with the other methods, which all use `KryoPersistenc`.

src/test/java/de/retest/recheck/persistence/bin/KryoPersistenceTest.java
diba1013: That would be two versions, I don't think that is really worth.

beatngu13 (approved):

PR #527, opened 2019-11-22
Feature/switch to graaljs

enesoezel:


beatngu13:
Have you had a look at the dependency tree as @modulo11 suggested? Does it pull many dependencies? And how does this affect the JAR size? (Especially the fat JAR we provide as part of recheck-web.)

martin-v:
It should be also considered that this lib is primary for GraalJVM, details to consider: https://github.com/graalvm/graaljs/blob/master/docs/user/RunOnJDK.md

enesoezel:
Due to the dependencies recheck-web would have 25 Mb instead of 15 Mb.
Following dependencies will be added because of graal:

[INFO] |  +- org.graalvm.js:js:jar:19.3.0:runtime
[INFO] |  |  +- org.graalvm.regex:regex:jar:19.3.0:runtime
[INFO] |  |  +- org.graalvm.truffle:truffle-api:jar:19.3.0:runtime
[INFO] |  |  +- org.graalvm.sdk:graal-sdk:jar:19.3.0:compile
[INFO] |  |  +- org.ow2.asm:asm:jar:7.1:compile
[INFO] |  |  +- org.ow2.asm:asm-tree:jar:7.1:runtime
[INFO] |  |  +- org.ow2.asm:asm-analysis:jar:7.1:runtime
[INFO] |  |  +- org.ow2.asm:asm-commons:jar:7.1:runtime
[INFO] |  |  +- org.ow2.asm:asm-util:jar:7.1:runtime
[INFO] |  |  \- com.ibm.icu:icu4j:jar:64.2:runtime
[INFO] |  \- org.graalvm.js:js-scriptengine:jar:19.3.0:compile

beatngu13:
Hmmm, so 167% file size.

Maybe we want another POC with Rhino, just for comparison? What do you others think?

roesslerj:
I agree that we should try out other things. If there are no other options, ok ... but there seem to be plenty.

PR #530, opened 2019-11-26
Have public save and load methods, such that they can be used by Babelfish

roesslerj:
Only a refactoring, shouldn't change impl details.

### DIFF
--- a/src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
+++ b/src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
@@ -90,10 +90,10 @@ public class KryoPersistence<T extends Persistable> implements Persistence<T> {
 		final Path path = Paths.get( identifier );
 		final File file = path.toFile();
 		FileUtil.ensureFolder( path.toFile() );
-		try ( OutputStream outputStream = newOutputStream( path ) ) {
+		try {
 			log.debug( "Writing {} to {}. Do not write to same identifier or interrupt until done.", element,
 					identifier );
-			save( outputStream, element );
+			save( newOutputStream( path ), element );
 			log.debug( "Done writing {} to {}", element, identifier );
 		} catch ( final Throwable t ) {
 			log.error( "Error writing to file {}. Deleting what has been written to not leave corrupt file behind...",
@@ -104,26 +104,22 @@ public class KryoPersistence<T extends Persistable> implements Persistence<T> {
 	}
 
 	public void save( final OutputStream outputStream, final T element ) throws IOException {
-		final Output output = new Output( new LZ4FrameOutputStream( outputStream ) );
-		output.writeString( version );
-		kryo.writeClassAndObject( output, element );
-		output.close();
+		try ( Output output = new Output( new LZ4FrameOutputStream( outputStream ) ) ) {
+			output.writeString( version );
+			kryo.writeClassAndObject( output, element );
+		}
 	}
 
 	@Override
 	public T load( final URI identifier ) throws IOException {
 		final Path path = Paths.get( identifier );
-		try {
-			return load( newInputStream( path ), identifier );
-		} catch ( final IncompatibleReportVersionException | NoSuchFileException e ) {
-			throw e;
-		}
+		return load( newInputStream( path ), identifier );
 	}
 
 	@SuppressWarnings( "unchecked" )
 	public T load( final InputStream in, final URI identifier ) throws IOException {
 		String writerVersion = null;
-		try ( final Input input = new Input( new LZ4FrameInputStream( in ) ) ) {
+		try ( Input input = new Input( new LZ4FrameInputStream( in ) ) ) {
 			writerVersion = input.readString();
 			final T persistable = (T) kryo.readClassAndObject( input );
 			if ( !isCompatible( persistable ) ) {
--- a/src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
+++ b/src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
@@ -5,6 +5,8 @@ import static java.nio.file.Files.newOutputStream;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.URI;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
@@ -88,11 +90,10 @@ public class KryoPersistence<T extends Persistable> implements Persistence<T> {
 		final Path path = Paths.get( identifier );
 		final File file = path.toFile();
 		FileUtil.ensureFolder( path.toFile() );
-		try ( final Output output = new Output( new LZ4FrameOutputStream( newOutputStream( path ) ) ) ) {
-			output.writeString( version );
+		try ( OutputStream outputStream = newOutputStream( path ) ) {
 			log.debug( "Writing {} to {}. Do not write to same identifier or interrupt until done.", element,
 					identifier );
-			kryo.writeClassAndObject( output, element );
+			save( outputStream, element );
 			log.debug( "Done writing {} to {}", element, identifier );
 		} catch ( final Throwable t ) {
 			log.error( "Error writing to file {}. Deleting what has been written to not leave corrupt file behind...",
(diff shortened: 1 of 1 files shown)


src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -102,12 +103,27 @@ public void save( final URI identifier, final T element ) throws IOException {
...
 		final Path path = Paths.get( identifier );
+		try {
+			return load( newInputStream( path ), identifier );
+		} catch ( final IncompatibleReportVersionException | NoSuchFileException e ) {
+			throw e;
+		}
+	}
+
+	@SuppressWarnings( "unchecked" )
+	public T load( final InputStream in, final URI identifier ) throws IOException {
```
diba1013, line 120: Do you have `URI` with your stream? If not, I would try to remove the identifier (as it is only used in the exception). I would generally try to keep the methods the same:
```
T load( final URI identifier ) throws IOException;
T load( final InputStream identifier ) throws IOException;
```

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -88,11 +90,10 @@ public void save( final URI identifier, final T element ) throws IOException {
 		final Path path = Paths.get( identifier );
 		final File file = path.toFile();
 		FileUtil.ensureFolder( path.toFile() );
-		try ( final Output output = new Output( new LZ4FrameOutputStream( newOutputStream( path ) ) ) ) {
-			output.writeString( version );
+		try ( OutputStream outputStream = newOutputStream( path ) ) {
```
diba1013: Please use try-with-resources here, just pass in the OutputStream (similarly to `load`). You should always close the outer most stream only, which should take care of dependent streams.
```
try {
    return save( newOutputStream( path ), element );
} catch ( final Throwable t ) {
    ....
}
```

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -102,12 +103,27 @@ public void save( final URI identifier, final T element ) throws IOException {
 		}
 	}
 
-	@SuppressWarnings( "unchecked" )
+	public void save( final OutputStream outputStream, final T element ) throws IOException {
+		final Output output = new Output( new LZ4FrameOutputStream( outputStream ) );
+		output.writeString( version );
+		kryo.writeClassAndObject( output, element );
+		output.close();
```
diba1013: This output will not be closed on error, see above.

diba1013 (changes requested):

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
roesslerj: Yeah, I tried that ... but I think this is really helpful information in case of errors. So how could you remove that then?

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
roesslerj: On error, this is no problem, since the resulting file would be corrupt anyways. This close is not there to prevent memory leaks (which most often is the result of forgetting to call close), but to prevent a corrupt result file.

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
roesslerj: Hmm. I am sorry. Maybe I do not understand correctly. But this currently is try-with-resource, right?

roesslerj: Closing a stream multiple times is no problem. Closing an outer stream will also close an inner stream, but closing an inner stream does not close the outer stream. Therefore, close twice ... just to make sure.

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -102,12 +103,35 @@ public void save( final URI identifier, final T element ) throws IOException {
...
+
+	/**
+	 * You need to maintain the stream and are responsible for closing it properly.
+	 */
+	@SuppressWarnings( "unchecked" )
+	public T load( final InputStream in, final URI identifier ) throws IOException {
 		String writerVersion = null;
-		try ( final Input input = new Input( new LZ4FrameInputStream( newInputStream( path ) ) ) ) {
+		try {
+			final Input input = new Input( new LZ4FrameInputStream( in ) );
```
diba1013: Why did you change this? This was fine as it is.

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -102,12 +103,35 @@ public void save( final URI identifier, final T element ) throws IOException {
...
-	@SuppressWarnings( "unchecked" )
+	/**
+	 * You need to maintain the stream and are responsible for closing it properly.
+	 */
+	public void save( final OutputStream outputStream, final T element ) throws IOException {
+		final Output output = new Output( new LZ4FrameOutputStream( outputStream ) );
+		output.writeString( version );
+		kryo.writeClassAndObject( output, element );
+		// Flush the stream to avoid corrupt file data
+		output.flush();
```
diba1013: I thought we discussed this as try-with-resources? Just to close the outer stream at all times.
```
try (final Output output = new Output( new LZ4FrameOutputStream( outputStream ) ) ) {
    output.writeString( version );
    kryo.writeClassAndObject( output, element );
}
```

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -102,12 +103,35 @@ public void save( final URI identifier, final T element ) throws IOException {
...
+	}
+
 	@Override
 	public T load( final URI identifier ) throws IOException {
 		final Path path = Paths.get( identifier );
+		try ( InputStream input = newInputStream( path ) ) {
+			return load( input, identifier );
+		} catch ( final IncompatibleReportVersionException | NoSuchFileException e ) {
+			throw e;
+		}
```
diba1013: We can just let the exception fall through:
```
try ( InputStream input = newInputStream( path ) ) {
	return load( input, identifier );
}
```

diba1013 (changes requested):

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
roesslerj: No it was not. We _must_ close the outermost stream, not some stream "in between". Closing here would have done the latter...

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -102,12 +103,33 @@ public void save( final URI identifier, final T element ) throws IOException {
 		}
 	}
 
-	@SuppressWarnings( "unchecked" )
+	/**
+	 * You need to maintain the stream and are responsible for closing it properly.
+	 */
```
diba1013: This comment is not true anymore

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
```diff
@@ -102,12 +103,33 @@ public void save( final URI identifier, final T element ) throws IOException {
...
+		try ( InputStream input = newInputStream( path ) ) {
+			return load( input, identifier );
+		} catch ( final IncompatibleReportVersionException | NoSuchFileException e ) {
+			throw e;
+		}
+	}
+
+	/**
+	 * You need to maintain the stream and are responsible for closing it properly.
+	 */
```
diba1013: This comment is not true anymore

diba1013 (changes requested):

src/main/java/de/retest/recheck/persistence/bin/KryoPersistence.java
roesslerj: Changes were lost during rebase. How annoying. Readded them.

diba1013 (approved):

PR #531, opened 2019-11-26
Add meta-data to SutState

roesslerj:

### DIFF
--- a/src/main/java/de/retest/recheck/ui/descriptors/SutState.java
+++ b/src/main/java/de/retest/recheck/ui/descriptors/SutState.java
@@ -2,7 +2,10 @@ package de.retest.recheck.ui.descriptors;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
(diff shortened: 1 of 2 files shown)


src/main/java/de/retest/recheck/ui/descriptors/SutState.java
```diff
@@ -22,25 +25,42 @@
 	@XmlElement
 	private final List<RootElement> descriptors;
 
+	@XmlElement
+	private final Map<String, String> metaData;
```
diba1013, line 29: [Metadata](https://en.wikipedia.org/wiki/Metadata) is one word only:
```suggestion
	private final Map<String, String> metadata;
```

diba1013 (changes requested):
Please take into consideration [RET-1896](https://retest.atlassian.net/browse/RET-1896) and its releated tasks.

Things to note:

1. Should we use a higher abstraction level (e.g. separate class)

    1. Keep it simple: No.
    2. Give it some thought: Yes. Then we could for example enforce a naming schema with namespaces and such for key collisions. Secondly, we do not have these confusing unmodifiable Map stuff, since we control the methods in it.
        ```
        @ToString
        @EqualsAndHashCode
        public class Metadata {

            Map<String, String> metadata;

            public Metadata(Map<Key, String> metadata) {
                this.metadata = // convert and flat the metadata
            }

            String get( final Key key );
            String getOrDefault( final Key key, String defaultValue );
        }

        @ToString
        @EqualsAndHashCode
        class Key {

            @Getter
            String key;

            public Key(String namespace, String part, String... additionalParts) {
                this( /* concat those with 'namespace:part1.part2' */ )
            }
        }
        ```
        And then use those `Key` to access the value behind.
2. This should be compatible with the old Golden Masters. However:

    1. Should we just in case increase the persistence version for that?
    2. How does JAXB handle loading of the old Golden Masters. Is the `metadata = null`?

src/main/java/de/retest/recheck/ui/descriptors/SutState.java
roesslerj: Done

roesslerj:
Forgot to adapt the metadata retrieval...

PR #534, opened 2019-11-27
Feature/revise filter interface

varlog23:

## Description

Expose new `matches(Element, String)` method that helps to implement a filter for `RootElement` and `Element` classes without having to create a fake `AttributeDifference` instance.

### DIFF
--- a/src/main/java/de/retest/recheck/ignore/CacheFilter.java
+++ b/src/main/java/de/retest/recheck/ignore/CacheFilter.java
@@ -19,6 +19,7 @@ public class CacheFilter implements Filter {
 
 	private final Map<Element, Boolean> elementCache = new HashMap<>();
 	private final Map<Pair<Element, AttributeDifference>, Boolean> attributeCache = new HashMap<>();
+	private final Map<Pair<Element, String>, Boolean> attributeKeyCache = new HashMap<>();
 
 	@Getter
 	private final Filter base;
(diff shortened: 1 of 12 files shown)


src/main/java/de/retest/recheck/ignore/Filter.java
```diff
@@ -58,6 +75,11 @@ public boolean matches( final Element element ) {
 			return false;
 		}
 
+		@Override
+		public boolean matches( final Element element, final String attributeKey ) {
```
beatngu13: Since `matches( Element, AttributeDifference )` defaults to this method, and this method defaults to `matches( Element )`, isn't it sufficient to only overwrite the latter?

varlog23: You are correct that it is needless to override `matches( Element, String )` but the existing overridden `matches( Element, AttributeDifference )` needs to be retained to avoid `java.lang.NullPointerException` in the below scenario.

```
final AttributeDifference attributeDifference = null;
final Filter filter = Filter.FILTER_NOTHING;
filter.matches( element, attributeDifference );
```

beatngu13:
I'm good with this PR, @diba1013 will have another look.

src/main/java/de/retest/recheck/ignore/CacheFilter.java
```diff
@@ -37,10 +38,19 @@ public boolean matches( final Element element, final AttributeDifference attribu
 		return attributeCache.computeIfAbsent( Pair.of( element, attributeDifference ), matchesDiff() );
 	}
 
+	@Override
+	public boolean matches( final Element element, final String attributeKey ) {
+		return attributeKeyCache.computeIfAbsent( Pair.of( element, attributeKey ), matchesKeyDiff() );
+	}
+
```
diba1013, line 49: I would move this method below, so that each computation function is below its `matches` function.

diba1013 (changes requested):
Just a minor thing...

src/main/java/de/retest/recheck/ignore/Filter.java
diba1013: Well, I would hope that we can establish with new APIs that no one is allowed to pass `null` as a parameter, unless it is specified.

diba1013 (approved):

PR #535, opened 2019-11-27
Feature/switch to rhino

### LINKED ISSUE #331: Warning: Nashorn engine is planned to be removed from a future JDK release
I'm using JDK 11, so I assume this should need to be addressing soon.

enesoezel:
There's almost no difference in size regarding the recheck-web jars.
All IT-Test in recheck-web seem to work as before.

resolves #331

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -192,6 +192,18 @@
 			<version>1.8</version>
 		</dependency>
 
+		<dependency>
+			<groupId>org.mozilla</groupId>
+			<artifactId>rhino</artifactId>
+			<version>1.7.11</version>
+		</dependency>
+
+		<dependency>
+			<groupId>cat.inspiracio</groupId>
+			<artifactId>rhino-js-engine</artifactId>
+			<version>1.7.10</version>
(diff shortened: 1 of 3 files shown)


diba1013:
Since you needed to adapt a test, is there any other compatibility problems with rhino? 
1. Does a simple cast to string work?
1. Does the recheck-web ignore.js work?
1. ...

enesoezel:
As I mentioned in the description it seems like recheck-web is working proper. The ignore file works also. Regarding [this stackoverflow post](https://stackoverflow.com/a/26509119), a simple cast does also work like: `String(diff.expected);` or `(diff.expected + "")`. I have tested both. It's working.

beatngu13:
So, too me it looks like this variant is superior to #527. Any objections against merging and testing it during dev and especially when releasing 1.8.0?

roesslerj:
I am also in favor of merging now and testing throughout the sprint.

roesslerj (approved):

beatngu13 (approved):

PR #537, opened 2019-11-27
Feature/add meta data

roesslerj:

### DIFF
--- a/src/main/java/de/retest/recheck/RecheckAdapter.java
+++ b/src/main/java/de/retest/recheck/RecheckAdapter.java
@@ -24,21 +24,21 @@ public interface RecheckAdapter {
 	/**
 	 * Returns {@code true} if the given object can be converted by the adapter.
 	 *
-	 * @param toVerify
-	 *            the object to verify
+	 * @param toCheck
+	 *            the object to check
 	 * @return true if the given object can be converted by the adapter
 	 */
-	boolean canCheck( Object toVerify );
+	boolean canCheck( Object toCheck );
 
 	/**
 	 * Convert the given object into a {@code RootElement} (respectively into a set of {@code RootElement}s if this is
 	 * sensible for this type of object).
 	 *
-	 * @param toVerify
-	 *            the object to verify
+	 * @param toCheck
+	 *            the object to check
 	 * @return The RootElement(s) for the given object
 	 */
-	Set<RootElement> convert( Object toVerify );
+	Set<RootElement> convert( Object toCheck );
 
 	/**
 	 * Returns a {@code DefaultValueFinder} for the converted element attributes. Default values of attributes are
--- a/src/main/java/de/retest/recheck/RecheckAdapter.java
+++ b/src/main/java/de/retest/recheck/RecheckAdapter.java
@@ -1,5 +1,7 @@
 package de.retest.recheck;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 
 import de.retest.recheck.report.ActionReplayResult;
@@ -39,6 +41,19 @@ public interface RecheckAdapter {
 	 * @return The RootElement(s) for the given object
 	 */
 	Set<RootElement> convert( Object toCheck );
+
+	/**
+	 * Return some metadata with respect to the checked object. For e.g. a selenium driver, this could be things like
+	 * browser name and version.
+	 *
+	 * @param toCheck
+	 *            the object to check
+	 *
+	 * @return The meta data for the given object
+	 */
+	default Map<String, String> retrieveMetadata( final Object toCheck ) {
+		return new HashMap<>();
+	}
 
 	/**
 	 * Returns a {@code DefaultValueFinder} for the converted element attributes. Default values of attributes are
--- a/src/main/java/de/retest/recheck/RecheckAdapter.java
+++ b/src/main/java/de/retest/recheck/RecheckAdapter.java
@@ -1,6 +1,6 @@
 package de.retest.recheck;
 
-import java.util.HashMap;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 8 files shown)


src/main/java/de/retest/recheck/RecheckAdapter.java
```diff
@@ -24,21 +26,34 @@ default RecheckAdapter initialize( final RecheckOptions opts ) {
...
+	 * Return some metadata with respect to the checked object. For e.g. a selenium driver, this could be things like
+	 * browser name and version.
+	 *
+	 * @param toCheck
+	 *            the object to check
+	 *
+	 * @return The meta data for the given object
+	 */
+	default Map<String, String> retrieveMetadata( final Object toCheck ) {
+		return new HashMap<>();
```
diba1013: I would use `Collections.emptyMap()` here, since we do not need mutability.

src/main/java/de/retest/recheck/ui/descriptors/SutState.java
```diff
@@ -78,6 +98,8 @@ public SutState applyChanges( final ActionChangeSet actionChangeSet ) {
 				descriptors.add( (RootElement) element );
 			}
 		}
-		return new SutState( descriptors );
+		final Map<String, String> newMetadata = new HashMap<>( metadata );
+		newMetadata.putAll( actionChangeSet.getMetadata() );
+		return new SutState( descriptors, newMetadata );
```
diba1013: It depends on how the implementation of `ActionChangeSet` handles the metadata. I would argue, that it contains the already merged (i.e. new) metadata (similar to the screenshot, we do no logic here). How is it supposed to delete metadata?

That means, following would suffice:
```
return new SutState( descriptor, actionChangeSet.getMetadata() )
```

src/main/java/de/retest/recheck/ui/review/ActionChangeSet.java
```diff
@@ -14,16 +16,23 @@
 	private final String description;
 	private final String goldenMasterPath;
 	private final ScreenshotChanges screenshots;
+	private final Map<String, String> metadata;
 
 	public ActionChangeSet() {
-		this( null, null, ScreenshotChanges.empty() );
+		this( null, null, ScreenshotChanges.empty(), new HashMap<>() );
```
diba1013: I would always prefer `Collections.emptyMap()` for such cases to avoid creating multiple instances.

src/main/java/de/retest/recheck/ui/review/ActionChangeSet.java
```diff
@@ -14,16 +16,23 @@
...
 	}
 
 	public ActionChangeSet( final String description, final String goldenMasterPath,
 			final ScreenshotChanges screenshots ) {
+		this( description, goldenMasterPath, screenshots, new HashMap<>() );
```
diba1013: I would always prefer `Collections.emptyMap()` for such cases to avoid creating multiple instances.

src/main/java/de/retest/recheck/ui/descriptors/SutState.java
```diff
@@ -22,25 +25,42 @@
...
 	@SuppressWarnings( "unused" )
 	private SutState() {
 		super( PERSISTENCE_VERSION );
 		// for JAXB
 		descriptors = new ArrayList<>();
+		metadata = new HashMap<>();
 	}
 
 	public SutState( final Collection<RootElement> set ) {
+		this( set, new HashMap<>() );
```
diba1013: I would always prefer `Collections.emptyMap()` for such cases to avoid creating multiple instances.

src/main/java/de/retest/recheck/ui/descriptors/SutState.java
```diff
@@ -22,25 +25,42 @@
...
 
+	@XmlElement
+	private final Map<String, String> metadata;
+
 	@SuppressWarnings( "unused" )
 	private SutState() {
 		super( PERSISTENCE_VERSION );
 		// for JAXB
 		descriptors = new ArrayList<>();
+		metadata = new HashMap<>();
```
diba1013, line 36: I would always prefer `Collections.emptyMap()` for such cases to avoid creating multiple instances.

diba1013 (changes requested):

src/main/java/de/retest/recheck/ui/descriptors/SutState.java
roesslerj: Can't do that here, as I am not sure how Jaxb handles deserialization... An empty Map can't be added elements to.

src/main/java/de/retest/recheck/RecheckAdapter.java
roesslerj: An empty map can't be added elements to. Since we do not ensure or enforce immutability of this result, we can't use empty map here.

diba1013: It is not possible to call default methods from overwritten methods, so only the result counts. We do not add elements to this, thus we do not need mutability after `retrieveMetadata` is called. I do not understand your logic here.

src/main/java/de/retest/recheck/ui/descriptors/SutState.java
```diff
@@ -22,25 +25,42 @@
 	@XmlElement
 	private final List<RootElement> descriptors;
 
+	@XmlElement
+	private final Map<String, String> metadata;
```
diba1013, line 29: I think that my comment got lost:

Since this changes the model, do we need to increase the `persistenceVersion` just in case? I think this should be compatible with old Golden Masters, but just to be sure that we could include the corresponding transformers for that in case.

We could do that separate if we realized that there are in fact incompatibilities.

diba1013 (changes requested):

src/main/java/de/retest/recheck/ui/descriptors/SutState.java
roesslerj: This should be a backwards compatible change (added field). Alsot, otherwise tests should have failed.

diba1013 (approved):

PR #538, opened 2019-11-28
Add RecheckProperties using owner

Creepow:

- [ ] the code follows the [Clean Code](https://clean-code-developer.com/) guidelines.
- [ ] all status checks (Travis CI, SonarCloud, etc.) pass.
- [ ] your commit history is cleaned up.

## Description
Draft PR to decide whether or not we use owner to manage our properties.
> TL;DR


diba1013:
Can this handle public and private properties? 

We have the minority of our system properties "exposed" to the user, but use several system properties internally. The private properties are just floating around, while the public ones are those that are written into the file. Can this library use such a type safe property retrieval for the private properties too, without writing that into the file specified? Or do we need an additional `@Sources` or interface for that?

I would imagine this to be the new `Properties` class where we collect all of our properties and then we can successively promote them firstly to public ones or into the `RecheckOptions`.

Otherwise we now have 3+ configuration points:

1. `Recheck(Web)Options`.
2. `RecheckProperties`.
3. `Properties`.
4. All the scattered internal system properties.

### Side note

I think @Creepow mentioned that you will explain this more in depth on monday. Does it read both the properties file and the system properties into this interface? What if I start this not in the project root, but instead need to search for the project (e.g. recheck.cli)?

E.g. I define following upon startup:
1. JVM options:
    ```
    -Dde.retest.recheck.ignore.attributes=foo;bar 
    ```
2. Properties file in project
    ```
    de.retest.recheck.ignore.attributes=hello;world
    ```
3. Properties file in home dir?
4. default.properties

It would suffice, if you could explain this or a few other scenarios on monday.

beatngu13:
@diba1013 we will go through the basics probably on Tuesday since Jeremy and Joe aren't available on Monday. And since we all have to use properties eventually, I would prefer the majority of the team to be present.

Regarding your actual questions: First of all, please note we are not looking for a silver bullet, but an improvement in terms of ease of use and less maintenance effort. My impression is that what I have seen so far seems already considerably better compared to what we have now.

>Can this handle public and private properties?

Not exactly sure what you mean by this, maybe we can have quick chat tomorrow so I can prepare an answer.

>Does it read both the properties file and the system properties into this interface?

Yes, we configured OWNER to do so. And we are able to define an order via `@Source` like:

```java
@Sources({ "file:.retest/retest.properties", 
           "file:~/.retest/retest.properties", 
           "classpath:default-retest.properties",
           "system:properties",
           "system:env" })
```

>What if I start this not in the project root, but instead need to search for the project (e.g. recheck.cli)?

~~I believe we have to configure this separately: http://owner.aeonbits.org/docs/configuring/~~

Done: https://github.com/retest/recheck/pull/538/files#diff-d9ee62e0b9945c4e801fe7ab78fc4ba3R15

beatngu13:
The basic migration should be done.

One thing that has changed: OWNER treats enums as case sensitive. Previously one was able to use, e.g., `plain`. However, as this is only internal, I assume we can live with it? Otherwise we would have to create a separate converter.

src/main/java/de/retest/recheck/RecheckProperties.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.List;
+
+import org.aeonbits.owner.Config.Sources;
+import org.aeonbits.owner.ConfigCache;
+import org.aeonbits.owner.ConfigFactory;
+import org.aeonbits.owner.Mutable;
+
+import de.retest.recheck.configuration.ProjectRootFinderUtil;
+
+@Sources( "file:${projectroot}/.retest/retest.properties" )
```
diba1013, line 13: Shouldn't we also read from the system properties? Or is this done with L23?

src/main/java/de/retest/recheck/ui/TargetFinder.java
```diff
@@ -18,6 +18,8 @@
 
 	private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger( TargetFinder.class );
 
+	static final double ELEMENT_MATCH_THRESHOLD = RecheckProperties.getInstance().elementMatchThreshold();
```
diba1013, line 21: If we use static here, the properties might not yet be initialized. Thus only the default value will be used. Please check this and cleanup the other instances.

src/main/java/de/retest/recheck/FileOutputFormat.java
```diff
@@ -0,0 +1,10 @@
+package de.retest.recheck;
```
diba1013: Shouldn't this be in the persistence package?

src/main/java/de/retest/recheck/RecheckProperties.java
```diff
@@ -0,0 +1,84 @@
...
+public interface RecheckProperties extends Mutable {
+
+	// Usage.
+
+	static void init() {
+		ProjectRootFinderUtil.getProjectRoot()
+				.ifPresent( projectRoot -> ConfigFactory.setProperty( "projectroot", projectRoot.toString() ) );
```
diba1013: Just to be save
```suggestion
				.ifPresent( projectRoot -> ConfigFactory.setProperty( "projectroot", projectRoot.toAbsolutePath().toString() ) );
```

src/main/java/de/retest/recheck/Rehub.java
```diff
@@ -30,7 +29,7 @@ public static void init() {
 			if ( !auth.isAuthenticated( getRecheckApiKey() ) ) {
 				auth.login( new RehubAuthenticationHandler() );
 			}
-			System.setProperty( Properties.FILE_OUTPUT_FORMAT_PROPERTY, FileOutputFormat.CLOUD.toString() );
+			RecheckProperties.getInstance().setProperty( "de.retest.output.Format", FileOutputFormat.CLOUD.toString() );
```
diba1013: Should we use a constant here?

src/main/java/de/retest/recheck/RecheckProperties.java
```diff
@@ -0,0 +1,84 @@
...
+	@Key( "de.retest.recheck.rootElementContainedChildrenMatchThreshold" )
+	@DefaultValue( "0.5" )
+	double rootElementContainedChildrenMatchThreshold();
+
+	@Key( "de.retest.recheck.rehub.reportUploadEnabled" )
+	@DefaultValue( "false" )
+	boolean rehubReportUploadEnabled();
+
+	@Key( "de.retest.output.Format" )
+	FileOutputFormat fileOutputFormat();
```
diba1013, line 86: If I see this correctly, this is never supposed to be called? Should we mark this as such?

src/main/java/de/retest/recheck/RecheckProperties.java
```diff
@@ -0,0 +1,84 @@
...
+	public static final String SCREENSHOT_FOLDER_NAME = "screenshot";
+	public static final String RECHECK_FOLDER_NAME = "recheck";
+	public static final String DEFAULT_XML_FILE_NAME = "retest.xml";
+
+	public static final String RETEST_FOLDER_NAME = ".retest";
+	public static final String RETEST_PROPERTIES_FILE_NAME = "retest.properties";
+
+	public static final String GOLDEN_MASTER_FILE_EXTENSION = ".recheck";
+	public static final String TEST_REPORT_FILE_EXTENSION = ".report";
+	public static final String AGGREGATED_TEST_REPORT_FILE_NAME = "tests" + TEST_REPORT_FILE_EXTENSION;
```
diba1013: Strictly speaking, there does not need to be a `public` modifier here.

diba1013 (changes requested):

src/main/java/de/retest/recheck/Rehub.java
diba1013: After looking at recheck-web, I would suggest to do so.

src/main/java/de/retest/recheck/FileOutputFormat.java
beatngu13: It wasn't before, but I can move it.

src/main/java/de/retest/recheck/RecheckProperties.java
beatngu13: Yes, this is covered by L23: http://owner.aeonbits.org/docs/importing-properties/

src/main/java/de/retest/recheck/RecheckProperties.java
beatngu13: `Rehub` sets the value of this property if enabled. I would rather ask: Do we still need this general property, or should we have separate properties for the state and report format? That is essentially how it is used internally.

rebazer:

src/main/java/de/retest/recheck/RecheckProperties.java
diba1013: Valid point. But I would address this separately.

rebazer:

rebazer:

rebazer:

rebazer:

rebazer:

src/main/java/de/retest/recheck/ui/TargetFinder.java
beatngu13: Works as expected, also in other places. `init()` is called beforehand.

diba1013 (approved):

PR #539, opened 2019-11-28
Feature/filters to improve sut model

varlog23:

## Description

Adding `RootElementFilter`, a utility class that can filter `RootElement`, `Element`, `IdentifyingAttributes` and `Attributes` instance. This is developed to ensure `SutStateIdentifier` in surili can produce same hash for a given SUT state.

### DIFF
--- a/src/main/java/de/retest/recheck/ui/descriptors/SutStateFilter.java
+++ b/src/main/java/de/retest/recheck/ui/descriptors/SutStateFilter.java
@@ -1,0 +1,90 @@
+package de.retest.recheck.ui.descriptors;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import de.retest.recheck.ignore.Filter;
+import de.retest.recheck.ui.image.Screenshot;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public class SutStateFilter {
+	private final Filter filter;
+
+	public SutState filter( final SutState sutState ) {
+		if ( filterNothing() ) {
+			return sutState;
+		}
+		final List<RootElement> rootElements = sutState.getRootElements();
+		final List<RootElement> filteredRootElements = rootElements.stream() //
+				.map( this::filter ) //
+				.collect( Collectors.toList() );
+		final SutState newSutState = new SutState( filteredRootElements, sutState.getMetadata() );
+		return newSutState;
+	}
+
+	public RootElement filter( final RootElement rootElement ) {
+		if ( filterNothing() ) {
+			return rootElement;
+		}
+		final IdentifyingAttributes newIdentifyingAttributes =
+				filter( rootElement, rootElement.getIdentifyingAttributes() );
+		final Attributes newAttributes = filter( rootElement, rootElement.getAttributes() );
+		final String retestId = rootElement.getRetestId();
+		final Screenshot screenshot = rootElement.getScreenshot();
+		final String screen = rootElement.getScreen();
+		final int screenId = rootElement.getScreenId();
+		final String title = rootElement.getTitle();
+		final List<Element> containedElements = rootElement.getContainedElements();
+
+		final RootElement newRootElement = new RootElement( retestId, newIdentifyingAttributes, newAttributes,
+				screenshot, screen, screenId, title );
+
+		final List<Element> newContainedElements = filterContainedElements( newRootElement, containedElements );
+		newRootElement.addChildren( newContainedElements );
+		return newRootElement;
+	}
+
+	IdentifyingAttributes filter( final Element element, final IdentifyingAttributes identifyingAttributes ) {
+		final List<Attribute> attributes = identifyingAttributes.getAttributes();
+
+		final List<Attribute> filteredAttributes = attributes.stream() //
+				.filter( attribute -> !filter.matches( element, attribute.getKey() ) ) //
+				.collect( Collectors.toList() );
+		return new IdentifyingAttributes( filteredAttributes );
+	}
+
+	Attributes filter( final Element element, final Attributes attributes ) {
+		final MutableAttributes filteredAttributes = new MutableAttributes();
+
+		StreamSupport.stream( attributes.spliterator(), false ) //
+				.filter( attribute -> !filter.matches( element, attribute.getKey() ) ) //
+				.forEach( attribute -> filteredAttributes.put( attribute.getKey(), (String) attribute.getValue() ) );
+
+		return filteredAttributes.immutable();
+	}
+
+	private Element filter( final Element element, final Element parent ) {
+		final IdentifyingAttributes newIdentifyingAttributes = filter( element, element.getIdentifyingAttributes() );
+		final Attributes newAttributes = filter( element, element.getAttributes() );
+		final String retestId = element.getRetestId();
+		final List<Element> containedElements = element.getContainedElements();
+
+		final Element newElement = Element.create( retestId, parent, newIdentifyingAttributes, newAttributes );
+
+		final List<Element> newContainedElements = filterContainedElements( newElement, containedElements );
+		newElement.addChildren( newContainedElements );
+		return newElement;
+	}
+
+	private List<Element> filterContainedElements( final Element newParent, final List<Element> containedElements ) {
+		return containedElements.stream() //
+				.map( element -> filter( element, newParent ) ) //
+				.collect( Collectors.toList() );
+	}
+
+	private boolean filterNothing() {
+		return filter == Filter.FILTER_NOTHING;
+	}
+}
--- a/src/test/java/de/retest/recheck/ui/descriptors/SutStateFilterTest.java
+++ b/src/test/java/de/retest/recheck/ui/descriptors/SutStateFilterTest.java
@@ -1,0 +1,303 @@
+package de.retest.recheck.ui.descriptors;
+
+import static de.retest.recheck.ui.Path.fromString;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+import de.retest.recheck.ignore.CompoundFilter;
+import de.retest.recheck.ignore.Filter;
+import de.retest.recheck.review.ignore.AttributeFilter;
+import de.retest.recheck.ui.Path;
+import de.retest.recheck.ui.image.Screenshot;
+import de.retest.recheck.ui.image.Screenshot.ImageType;
+
+class SutStateFilterTest {
+
+	@Nested
+	class AttributeFilterTests {
+		Element element;
+		Attributes attributes;
+
+		@BeforeEach
+		void setUp() {
+			element = mock( Element.class );
+			final MutableAttributes mutable = new MutableAttributes();
+			mutable.put( "enabled", "true" );
+			mutable.put( "rowCount", "5" );
+			attributes = mutable.immutable();
+		}
+
+		@Test
+		void should_return_same_attributes_with_filter_nothing() {
+			final Filter noFilter = Filter.FILTER_NOTHING;
+			final SutStateFilter cut = new SutStateFilter( noFilter );
+			final Attributes newState = cut.filter( element, attributes );
+
+			assertThat( (Object) newState ).isEqualTo( attributes );
+		}
+
+		@Test
+		void should_return_same_attributes_with_no_filter() {
+			final Filter noFilter = ( element ) -> false;
+			final SutStateFilter cut = new SutStateFilter( noFilter );
+			final Attributes newState = cut.filter( element, attributes );
+
+			assertThat( (Object) newState ).isEqualTo( attributes );
+		}
+
+		@Test
+		void should_filter_respective_attribute() {
+			final Filter filter = new AttributeFilter( Attributes.ENABLED );
+			final SutStateFilter cut = new SutStateFilter( filter );
+			final Attributes newState = cut.filter( element, attributes );
+
+			assertThat( newState.size() ).isEqualTo( 1 );
+			assertThat( newState.get( "enabled" ) ).isNull();
+			assertThat( newState.get( "rowCount" ) ).isEqualTo( "5" );
+		}
+
+		@Test
+		void should_not_do_anything_with_no_matches() {
+			final Filter filter = new AttributeFilter( Attributes.FONT_FAMILY );
+			final SutStateFilter cut = new SutStateFilter( filter );
+			final Attributes newState = cut.filter( element, attributes );
+
+			assertThat( (Object) newState ).isEqualTo( attributes );
+		}
+
+	}
+
+	@Nested
+	class IdentifyingAttributesFilterTests {
+		Element element;
+		IdentifyingAttributes identifyingAttributes;
+
+		@BeforeEach
+		void setUp() {
+			final Path path = Path.fromString( "Window[1]/Path[1]/Component[1]" );
+			element = mock( Element.class );
+			identifyingAttributes = IdentifyingAttributes.create( path, Component.class );
+		}
+
+		@Test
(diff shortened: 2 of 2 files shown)


src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.List;
+import java.util.Map.Entry;
+import java.util.stream.Collectors;
+
+import de.retest.recheck.ignore.Filter;
+import de.retest.recheck.ui.image.Screenshot;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public class RootElementFilter {
```
beatngu13: Maybe I didn't understand it correctly, but wasn't the initial idea to have an `applyFilter` method in `Element` and `RootElement`, just like `applyChanges`?

However, I believe I prefer this approach similar to `TestReportFilter` as this also removes complexity from the core classes.

varlog23: I initially implemented an `applyFilter` method in respective classes. However, @diba1013 also preferred this approach for the exact same reason you mentioned. `RootElementFilter` probably isn't the best name for a utility class. And I would rename it if you come up with a suggestion.

beatngu13: Personally I'm fine with the name. It is similar to `TestReportFilter`, which starts with the top container. The only thing we could consider is to have a `SutStateFilter` instead.

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
```diff
@@ -0,0 +1,96 @@
...
+	private final Filter filter;
+
+	public RootElement filter( final RootElement rootElement ) {
+		if ( filter.equals( Filter.FILTER_NOTHING ) ) {
```
beatngu13: Maybe extract a separate method `filterNothing()`?

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
```diff
@@ -0,0 +1,96 @@
...
+		return newRootElement;
+	}
+
+	public IdentifyingAttributes filter( final Element element, final IdentifyingAttributes identifyingAttributes ) {
+		if ( filter.equals( Filter.FILTER_NOTHING ) ) {
+			return identifyingAttributes;
+		}
+		final List<Attribute> attributes = identifyingAttributes.getAttributes();
+
+		final Collection<Attribute> filteredAttributes = attributes.stream() //
```
beatngu13: I would use a `List` as in `attributes`.

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
```diff
@@ -0,0 +1,96 @@
...
+				.collect( Collectors.toList() );
+		return new IdentifyingAttributes( filteredAttributes );
+	}
+
+	public Attributes filter( final Element element, final Attributes attributes ) {
+		if ( filter.equals( Filter.FILTER_NOTHING ) ) {
+			return attributes;
+		}
+		final MutableAttributes filteredAttributes = new MutableAttributes();
+		final Iterator<Entry<String, Object>> iterator = attributes.iterator();
```
beatngu13: Instead of an iterator, you can a for loop:

```java
for ( final Entry<String, Object> attribute : attributes ) {
	if ( filter.matches( element, attribute.getKey() ) ) {
		continue;
	}
	filteredAttributes.put( attribute.getKey(), (String) attribute.getValue() );
}
```

Or like elsewhere with a stream:

```java
StreamSupport.stream( attributes.spliterator(), false ) //
		.filter( attribute -> !filter.matches( element, attribute.getKey() ) ) //
		.forEach( attribute -> filteredAttributes.put( attribute.getKey(), (String) attribute.getValue() ) );
);
```

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
```diff
@@ -0,0 +1,96 @@
...
+
+		final Element newElement = Element.create( retestId, parent, newIdentifyingAttributes, newAttributes );
+
+		final List<Element> newContainedElements = filterContainedElements( newElement, containedElements );
+		newElement.addChildren( newContainedElements );
+		return newElement;
+	}
+
+	private List<Element> filterContainedElements( final Element newParent, final List<Element> containedElements ) {
+		final List<Element> newContainedElements = containedElements.stream() //
```
beatngu13: You can inline the variable:

```java
return containedElements.stream() //
		.map( element -> filter( element, newParent ) ) //
		.collect( Collectors.toList() );
```

src/test/java/de/retest/recheck/ui/descriptors/RootElementFilterTest.java
```diff
@@ -0,0 +1,314 @@
...
+
+class RootElementFilterTest {
+
+	RootElementFilter rootElementFilter;
+
+	@Nested
+	class AttributeFilterTests {
+		private final Element element = mock( Element.class );
+		Filter filter;
+		Attributes state;
```
beatngu13: Hmmm, I find "state" confusing. Is there a reason why you didn't name it "attributes"?

src/test/java/de/retest/recheck/ui/descriptors/RootElementFilterTest.java
```diff
@@ -0,0 +1,314 @@
...
+import de.retest.recheck.ui.image.Screenshot;
+import de.retest.recheck.ui.image.Screenshot.ImageType;
+
+class RootElementFilterTest {
+
+	RootElementFilter rootElementFilter;
+
+	@Nested
+	class AttributeFilterTests {
+		private final Element element = mock( Element.class );
```
beatngu13: I would stick to the given style and decouple it from the test class lifecycle, i.e., put it into the setup. Same elsewhere.

src/test/java/de/retest/recheck/ui/descriptors/RootElementFilterTest.java
```diff
@@ -0,0 +1,314 @@
...
+		Filter filter;
```
beatngu13: If each test uses its own filter and you don't set it up globally, remove the field. Same goes for `rootElementFilter` and others.

src/test/java/de/retest/recheck/ui/descriptors/RootElementFilterTest.java
```diff
@@ -0,0 +1,314 @@
...
+			final MutableAttributes mutable = new MutableAttributes();
+			mutable.put( "enabled", "true" );
+			mutable.put( "rowCount", "5" );
+			state = mutable.immutable();
+		}
+
+		@Test
+		public void should_return_same_attributes_with_filter_nothing() {
+			filter = Filter.FILTER_NOTHING;
+			rootElementFilter = new RootElementFilter( filter );
```
beatngu13: This should be named "cut" everywhere.

src/test/java/de/retest/recheck/ui/descriptors/RootElementFilterTest.java
```diff
@@ -0,0 +1,314 @@
...
+		@BeforeEach
+		void setUp() {
+			final MutableAttributes mutable = new MutableAttributes();
+			mutable.put( "enabled", "true" );
+			mutable.put( "rowCount", "5" );
+			state = mutable.immutable();
+		}
+
+		@Test
+		public void should_return_same_attributes_with_filter_nothing() {
```
beatngu13: It is JUnit 5, so all methods can be package private.

src/test/java/de/retest/recheck/ui/descriptors/RootElementFilterTest.java
```diff
@@ -0,0 +1,314 @@
...
+			filter = Filter.FILTER_NOTHING;
+			rootElementFilter = new RootElementFilter( filter );
+			final Attributes newState = rootElementFilter.filter( element, state );
+
+			assertThat( (Object) newState ).isEqualTo( state );
+		}
+
+		@Test
+		public void should_return_same_attributes_with_no_filter() {
+			filter = new CompoundFilter();
```
beatngu13: Maybe this is more self-explanatory compared to an empty compound:

```java
Filter noFilter = ( element ) -> false;
```

src/test/java/de/retest/recheck/ui/descriptors/RootElementFilterTest.java
```diff
@@ -0,0 +1,314 @@
...
+			rootElementFilter = new RootElementFilter( filter );
+			final Attributes newState = rootElementFilter.filter( element, state );
+
+			assertThat( (Object) newState ).isEqualTo( state );
+		}
+
+		@Test
+		public void should_filter_respective_attribute() {
+			filter = new CompoundFilter( new AttributeFilter( Attributes.ENABLED ) );
```
beatngu13: Why do you wrap this (and others) into an additional compound?

src/test/java/de/retest/recheck/ui/descriptors/RootElementFilterTest.java
```diff
@@ -0,0 +1,314 @@
...
+			final Attributes newState = rootElementFilter.filter( element, state );
+
+			assertThat( (Object) newState ).isEqualTo( state );
+		}
+
+	}
+
+	@Nested
+	class IdentifyingAttributesFilterTests {
+		private final Path path = Path.fromString( "Window[1]/path[1]/component[1]" );
```
beatngu13: Here `Window[1]/path[1]/component[1]`, below `Window[1]/Parent[1]/Component[2]`. Try to be more consistent.

src/test/java/de/retest/recheck/ui/descriptors/RootElementFilterTest.java
```diff
@@ -0,0 +1,314 @@
...
+import org.junit.jupiter.api.Test;
+
+import de.retest.recheck.ignore.CompoundFilter;
+import de.retest.recheck.ignore.Filter;
+import de.retest.recheck.review.ignore.AttributeFilter;
+import de.retest.recheck.ui.Path;
+import de.retest.recheck.ui.image.Screenshot;
+import de.retest.recheck.ui.image.Screenshot.ImageType;
+
+class RootElementFilterTest {
```
beatngu13: Did you copy some test code, especially `RootElementFilterTests` and below? Some parts contain a few legacy names etc., which we should clean up when we add new code.

beatngu13 (changes requested):
We can also discuss this tomorrow if you want to. ✌️

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
```diff
@@ -0,0 +1,96 @@
...
+			final Entry<String, Object> entry = iterator.next();
+			if ( filter.matches( element, entry.getKey() ) ) {
+				continue;
+			}
+			filteredAttributes.put( entry.getKey(), (String) entry.getValue() );
+		}
+		return filteredAttributes.immutable();
+	}
+
+	public Element filter( final Element element ) {
```
diba1013: Do you really need to filter an single element? Because to be complete, you would need to replace the new child within the parent at the same index once done. Otherwise you would have the new child floating around with only the connection to the parent, but the parent does not now about its child. That would be a sad story 😢.

diba1013 (dismissed):
Please make sure that only the methods that you really need are public. If you need to test the other methods, make them package private. So I think the public methods worthy to be public would be (you may choose to only have a subset of those, i.e. only `SutState` or only `RootElement`):

* `filter( SutState goldenMaster )`
* `filter( RootElement )`
* filter( Element )` only if really necessary, since it is more complex (see comment below)

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
varlog23: Removing this check would mean a new `RootElement` instance is created with `FILTER_NOTHING`. I believe extracting this logic to a separate method is a better idea

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
varlog23: That would be a sad story indeed. But doesn't line 80 ensures that this situation never occurs?

`newElement.addChildren( newContainedElements );`

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
```diff
@@ -0,0 +1,96 @@
...
+		return newElement;
+	}
+
+	private Element filter( final Element element, final Element parent ) {
+		final IdentifyingAttributes newIdentifyingAttributes = filter( element, element.getIdentifyingAttributes() );
+		final Attributes newAttributes = filter( element, element.getAttributes() );
+		final String retestId = element.getRetestId();
+		final List<Element> containedElements = element.getContainedElements();
+
+		final Element newElement = Element.create( retestId, parent, newIdentifyingAttributes, newAttributes );
```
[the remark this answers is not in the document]
varlog23: Yes, I did. I've added tests to verify the parents are updated from bottom to top.

src/test/java/de/retest/recheck/ui/descriptors/RootElementFilterTest.java
varlog23: Yes, I did. I did not realize this at the time. I will clean this up.

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
diba1013: But that would only be the case for the grand children. The `element`s parent does not get updated, because [L67](https://github.com/retest/recheck/blob/9b2503a16ec5ff991fd3223199f61859970f127c/src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java#L67) only creates the new child. 

You could argue, that it is the task of the caller to update the method. But on the other hand, is this method really needed at the moment?

varlog23: You are absolutely right. I completely missed that point. As we don't need this method at the moment, I would agree that this needs to go.

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
beatngu13: In the case of surili, we have to filter entire SUT states. This would also correspond to the `TestReportFilter`, i.e., the top container. We can still offer public methods for `RootElement`, `Element`, etc.

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
```diff
@@ -0,0 +1,93 @@
...
+
+import de.retest.recheck.ignore.Filter;
+import de.retest.recheck.ui.image.Screenshot;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public class RootElementFilter {
+	private final Filter filter;
+
+	public SutState filter( final SutState goldenMaster ) {
```
beatngu13: A SUT state is not necessarily a golden master, whereas a golden master is always a SUT state. Consider expected and actual. Both are SUT state, but only expected has been approved by a human, so only this is the golden master.

```suggestion
	public SutState filter( final SutState sutState ) {
```

src/test/java/de/retest/recheck/ui/descriptors/RootElementFilterTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+	@Nested
+	class AttributeFilterTests {
+		private Element element;
```
beatngu13: Why is this `element` private but `attributes` not? We don't expect test fields to be private, but either way pick one variant.

beatngu13 (changes requested):

src/main/java/de/retest/recheck/ui/descriptors/RootElementFilter.java
varlog23: We will offer public methods for `SutState` and `RootElement` for now.

beatngu13 (approved):

PR #547, opened 2019-12-03
Feature/let metadata overwrite others

diba1013:

## Description

> TL;DR Allow merges for `MetadataProvider`. This is only enabled for the `MetadataProviderService` as this allows for the global and local metadata.

Per documentation the `MetadataProvider` is now allowed to allow for duplicate keys by providing a respective merge strategy. This is implemented by the `MultiMetadataProvider` to allow for such a merge strategy to be used. However, the default behavior is still to disallow the merging and throw an exception. That is so that the decision must be willingly taken to allow merges as it conforms to the `Collectors#toMap` functions.

This is further implemented:
1. `GlobalMetadataProvider` should disallow duplicate keys in order to keep the global metadata providers to cancel each other out.
2. `MetadataProviderService` allows duplicate keys by explicitly allowing the local metadata to take priority.

### DIFF
--- a/src/main/java/de/retest/recheck/meta/MetadataProvider.java
+++ b/src/main/java/de/retest/recheck/meta/MetadataProvider.java
@@ -23,9 +23,6 @@ import java.util.Map;
  * <li><b>No {@code null} values are allowed.</b> Keys that map to non existent values (i.e. {@code null} should not be
  * added as this may be used as indicator for a missing key. Instead an other representation (e.g. {@code ""}) should be
  * used.</li>
- * <li><b>No duplicate keys are allowed.</b> A provider is not permitted to overwrite or alter any metadata of another
- * provider. Precautions should be taken to avoid (unintended) duplicate keys, such as a namespace for each
- * provider.</li>
  * <li><b>Only simple values shall be used.</b> While theoretically any arbitrary data may be provided, it should be
  * taken care, that each datum is as simple as possible. It should therefore be avoided to parse complex data from a
  * single datum. Instead, multiple correlated keys shall be used and then constructed as necessary by retrieving the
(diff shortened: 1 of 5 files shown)


roesslerj:
Hmm. Das halte ich für sehr kompliziert. M.E. gibt es keinen sinnvollen Grund für Throw.

diba1013:
`MergeStrategies.THROW` just mimics the behavior from `Collectors.toMap(Function, Function)` where no merge function is specified. This is for a provider to disallow duplicate keys (as specified by the documentation). E.g. we do not want the `GlobalMetadataProvider` to allow for duplicate keys, since each global provider should provide unique keys and not overwrite the already existing keys.

As stated above, I would prefer to mimic the behavior of `Collectors.toMap` where the merge function must be specified explicitly in order to allow for duplicate keys. This behavior is also reflected in the documentation provided. A so called *merging provider* must specify a merging strategy, otherwise it must be assumed that no duplicate keys are allowed.

The real change is, as we discussed, in the `MetadataProviderService` where it is explicitly specified that the local metadata should take precedence. Therefore it is clear and well defined which merging strategy is used instead of relying and searching for the default behavior.

roesslerj:
I think we shouldn't implement functionality just because we can. I do not see a use case for a THROW strategy. Can we just implement this such that keys get overwritten and a warning logged? All else is IMHO over-engineered for a mere >10 keys and a use case that's unsure when it will occur if ever.

diba1013:
I added a test for the `GlobalMetadataProvider` to make my case a bit clearer. Originally I indented for this to be part of #548.

My point with the exception is that it is much easier to guard against, than to just look for a log message. And for me, this test or, more specifically, this implementation then is not just there and it might not happen, but to notify us if it does happen, so that we are properly informed and can take actions.

roesslerj (approved):

PR #549, opened 2019-12-04
Feature/refactor auth handler

modulo11:
Second part of the basic auth refactoring.

### DIFF
--- a/src/main/java/de/retest/recheck/auth/AuthenticationHandler.java
+++ b/src/main/java/de/retest/recheck/auth/AuthenticationHandler.java
@@ -6,8 +6,8 @@ public interface AuthenticationHandler {
 
 	void showWebLoginUri( URI loginUri );
 
-	void authenticated();
+	void loginPerformed();
 
-	void authenticationFailed( Throwable reason );
+	void loginFailed( Throwable reason );
 
 }
--- a/src/main/java/de/retest/recheck/auth/AuthenticationHandler.java
+++ b/src/main/java/de/retest/recheck/auth/AuthenticationHandler.java
@@ -10,4 +10,6 @@ public interface AuthenticationHandler {
 
 	void loginFailed( Throwable reason );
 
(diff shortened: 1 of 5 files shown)


src/main/java/de/retest/recheck/auth/AuthenticationHandler.java
```diff
@@ -6,8 +6,10 @@
 
 	void showWebLoginUri( URI loginUri );
 
-	void authenticated();
+	void loginPerformed( String token );
```
roesslerj, line 9: Do you have a branch in place for that in downstream projects (read review)?

src/main/java/de/retest/recheck/auth/RetestAuthentication.java
```diff
@@ -37,7 +37,7 @@
 	public static final String AUTH_SERVER_PROPERTY = "de.retest.auth.server";
 	private static final String AUTH_SERVER_PROPERTY_DEFAULT = "https://sso.prod.cloud.retest.org/auth";
 	public static final String RESOURCE_PROPERTY = "de.retest.auth.resource";
-	private static final String RESOURCE_PROPERTY_DEFAULT = "review";
+	private static final String RESOURCE_PROPERTY_DEFAULT = "marvin";
```
roesslerj, line 40: Do we make sure in downstream projects to change that (i.e. change the default)?

src/main/java/de/retest/recheck/auth/AuthenticationHandler.java
modulo11: Yes, I do.

src/main/java/de/retest/recheck/auth/RetestAuthentication.java
modulo11: This is/can be overwritten (even in the `Rehub` class).

roesslerj (approved):

PR #551, opened 2019-12-04
Add GIT metadata

roesslerj:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Table of Contents
     * Operating system name and version.
     * Machine name (if available).
     * Date and time.
+    * Git branch and commit.
 
 ### Improvements
 
--- a/src/main/java/de/retest/recheck/meta/global/GitMetadataProvider.java
+++ b/src/main/java/de/retest/recheck/meta/global/GitMetadataProvider.java
@@ -1,0 +1,40 @@
+package de.retest.recheck.meta.global;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import de.retest.recheck.meta.MetadataProvider;
+import de.retest.recheck.persistence.GitExecutor;
+
+public class GitMetadataProvider implements MetadataProvider {
+
+	public static final String VCS_GIT = "git";
+	public static final String VCS_NAME = "vcs.name";
+	public static final String BRANCH_NAME = "vcs.branch";
+	public static final String COMMIT_NAME = "vcs.commit";
+
+	private final GitExecutor git;
+
+	public GitMetadataProvider() {
+		this( new GitExecutor() );
+	}
+
+	protected GitMetadataProvider( final GitExecutor git ) {
+		this.git = git;
+	}
+
+	@Override
+	public Map<String, String> retrieve() {
+		final Map<String, String> map = new HashMap<>();
+
+		final String branch = git.getCurrentBranch();
+		final String commit = git.getCurrentCommit();
+		if ( branch != null && commit != null ) {
+			map.put( VCS_NAME, VCS_GIT );
+			map.put( BRANCH_NAME, branch );
+			map.put( COMMIT_NAME, commit );
+		}
+
+		return map;
+	}
+}
--- a/src/main/java/de/retest/recheck/persistence/GitExecutor.java
+++ b/src/main/java/de/retest/recheck/persistence/GitExecutor.java
@@ -32,11 +32,16 @@ public class GitExecutor {
 
 	protected String executeGitCommand( final String command ) throws IOException, InterruptedException {
 		final Process process = Runtime.getRuntime().exec( command );
-		process.waitFor();
+		try {
+			process.waitFor();
 
-		final BufferedReader reader = new BufferedReader( new InputStreamReader( process.getInputStream() ) );
+			final BufferedReader reader = new BufferedReader( new InputStreamReader( process.getInputStream() ) );
 
-		return reader.readLine();
+			return reader.readLine();
+		} finally {
+			process.getInputStream().close();
+			process.destroy();
+		}
 	}
 
 	public String getCurrentBranch() {
--- a/src/main/java/de/retest/recheck/persistence/GitExecutor.java
+++ b/src/main/java/de/retest/recheck/persistence/GitExecutor.java
@@ -1,0 +1,71 @@
+package de.retest.recheck.persistence;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class GitExecutor {
+
+	protected static final String GIT_COMMAND_VERSION = "git --version";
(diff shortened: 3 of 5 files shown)


src/main/java/de/retest/recheck/persistence/GitUtil.java
```diff
@@ -0,0 +1,67 @@
...
+		try {
+			return executeGitCommand( GIT_COMMAND_VERSION ).startsWith( "git" );
+		} catch ( final Exception e ) {
+			log.debug( "`{}` resulted in exception {}, assuming no git installed.", GIT_COMMAND_VERSION,
+					e.getMessage() );
+			return false;
+		}
+	}
+
+	protected String executeGitCommand( final String command ) throws IOException, InterruptedException {
```
beatngu13: Just stumbled over this: rather than doing all of this manually, how about using JGit like we do in the rebazer?

Add this dependency:

```xml
<dependency>
	<groupId>org.eclipse.jgit</groupId>
	<artifactId>org.eclipse.jgit</artifactId>
	<version>5.5.1.201910021850-r</version>
</dependency>
```

This snippet should retrieve the name:

```java
final Repository repo = FileRepositoryBuilder.create( new File( "./.git/" ) );
try ( final Git git = new Git( repo ) ) {
	final ObjectId head = repo.resolve( "HEAD" );
	final Map<ObjectId, String> names = git.nameRev() //
			.add( head ) //
			.call();
	final String currentCommit = head.abbreviate( 7 ).name();
	final String currentBranchName = names.get( head );
	System.out.println( currentCommit + " on " + currentBranchName );
}
```

This should correspond to:

```
$ git name-rev HEAD
```

Maybe this can be done even more elegant?

https://stackoverflow.com/q/6245570

roesslerj: I don't like adding yet another dependency to replace a mere 4 lines of code.

roesslerj: As RET-1907 shows, the fewer dependencies, the better.

beatngu13: > I don't like adding yet another dependency to replace a mere 4 lines of code.

It replaces the content of the `GitUtil`, not just these 4 lines.

> As RET-1907 shows, the fewer dependencies, the better.

You definitely can't say that per se.

roesslerj: Hmm. I think "the fewer dependencies, the better" is something that generally holds. It's a always a tradeoff, so you usually get something in return. Just in this case, I think what you get is not worth the tradeoff.

src/main/java/de/retest/recheck/persistence/GitUtil.java
```diff
@@ -0,0 +1,67 @@
...
+
+	private final boolean localGit;
+
+	public GitUtil() {
+		localGit = isGitInstalled();
+	}
+
+	protected boolean isGitInstalled() {
+		try {
+			return executeGitCommand( GIT_COMMAND_VERSION ).startsWith( "git" );
```
diba1013: For Windows, the message is 
```
PS C:\Users\XXXX> git --version
git: The term 'git' is not recognized ...
```

diba1013 (changes requested):
I think, that these tests are kind of pointless, since you never actually execute git. I would assume, that writing a test for that is difficult, but I would at least create a IT test, with a temp directory and try to mock a environment so that you I have a proper way of creating a branch and checking the commit version.

src/main/java/de/retest/recheck/persistence/GitUtil.java
```diff
@@ -0,0 +1,67 @@
+package de.retest.recheck.persistence;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class GitUtil {
```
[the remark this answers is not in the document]
roesslerj: Ok

src/main/java/de/retest/recheck/persistence/GitUtil.java
roesslerj: Thanks, adapted accordingly.

src/main/java/de/retest/recheck/persistence/GitUtil.java
beatngu13: It's just a suggestion.

Also note that JGit is a pure Java implemention of Git. Suppose you have a dev that has no Git installed or not in his `PATH`—one can, for example, solely work with [EGit](https://www.eclipse.org/egit/)—JGit still works. In contrast, the current implementation doesn't as far as I can see.

diba1013: > Suppose you have a dev that has no Git installed or not in his PATH

That is a good point. This would then not just solve the manual problem (e.g. addressing this [comment](https://github.com/retest/recheck/pull/551#discussion_r353696258)), but make the execution independent of the installed git.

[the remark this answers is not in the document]
roesslerj: Added tests for that.

roesslerj: Hmm. I do not want to add 2.6 MB to replace a couple of lines of code. Currently I don't think we need a full-fledged git implementation. KISS.

beatngu13: I think the probability that a user encounters the aformentioned issues (no Git or not in `PATH`) is higher than the probability that a user complains about the (fat) JAR size or the number of downloaded dependencies. (Did you ever do that?)

And regarding KISS: I think this is actually a counter-example. To keep it simple, use something that "simply" works out-of-the-box. Why bother with a custom implementation, a local Git installation, OS differences, etc.

However, to end the discussion: your PR works, it is not a bad hack or something. I just think that using JGit has more pros than cons compared to the current implementation. So, it is your decision … but mark my words. 😉

roesslerj: In my opinion, it's not about the number of downloaded dependencies, but the potential problems that come with this (again see RET-1907) or the problems we have with one of our clients I mentioned... I'll mark your words, and I am open to replace that ones we really need that or the current solution causes problems. But until then, I'd like to keep it small...

src/main/java/de/retest/recheck/persistence/GitUtil.java
```diff
@@ -0,0 +1,67 @@
...
+
+	protected static final String GIT_COMMAND_VERSION = "git --version";
+	protected static final String GIT_COMMAND_BRANCH_NAME = "git rev-parse --abbrev-ref HEAD";
+	protected static final String GIT_COMMAND_COMMIT_HASH = "git rev-parse HEAD";
```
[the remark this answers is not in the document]
roesslerj: I have to admit that I don't recall how I came up with this. Stackoverflow is a good guess, but then whoever want's to know can also just google it ... like I would do now.

diba1013 (approved):

PR #555, opened 2019-12-05
Feature/fix xpath matcher

roesslerj:

### DIFF
--- a/src/main/java/de/retest/recheck/review/ignore/matcher/ElementXPathMatcher.java
+++ b/src/main/java/de/retest/recheck/review/ignore/matcher/ElementXPathMatcher.java
@@ -5,6 +5,7 @@ import java.util.regex.MatchResult;
 import java.util.regex.Pattern;
 
 import de.retest.recheck.review.ignore.io.RegexLoader;
+import de.retest.recheck.ui.Path;
 import de.retest.recheck.ui.descriptors.Element;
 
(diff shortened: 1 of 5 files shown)


beatngu13:
Cool, this partly adresses https://github.com/retest/recheck-web/issues/132 (see https://github.com/retest/recheck-web/issues/132#issuecomment-540779863).

src/main/java/de/retest/recheck/review/ignore/matcher/ElementXPathMatcher.java
```diff
@@ -26,7 +29,7 @@ public boolean test( final Element element ) {
 
 	@Override
 	public String toString() {
-		return String.format( ElementXpathMatcherLoader.FORMAT, xpath );
+		return String.format( ElementXpathMatcherLoader.FORMAT, "/" + xpath );
```
beatngu13: The leading slash is correct for the absolute XPath, but this doesn't correspond to our internal representation I guess? Is this good?

beatngu13 (changes requested):

src/main/java/de/retest/recheck/review/ignore/matcher/ElementXPathMatcher.java
roesslerj: I guess we should think about changing our internal representation. But this is definitely a separate PR...

beatngu13 (approved):

PR #556, opened 2019-12-06
Using the actual element is even _less_ code and more realistic

roesslerj:


beatngu13 (approved):
This is something we should consider in the future. Less mocking when it is possible to easily use "real" objects/data.

PR #558, opened 2019-12-06
Feature/migrate pixel diff filter

roesslerj:
-> Draft PR...

I would also create an e2e migration test, but this would be a copy of the test in https://github.com/retest/recheck/pull/557.


beatngu13:
Hmmm, I don't know. I personally don't like:

>Existing filters will be converted to `px`.

I think `pixel-diff` is already pretty clear; it's about pixels. So, why add the unit again? Until we support other variants (e.g. % as suggested by you), I would prefer to keep it as it is. Who knows what format we are going to use then.

beatngu13:
After yesterday's discussion, I believe this is a reasonable change. Would you mind updating the branch and marking it as ready for review?

roesslerj:
Absolutely not. But as mentioned:

> I would also create an e2e migration test, but this would be a copy of the test in #557.

So I'd wait for this branch to go in first, and then adapt the migration test.

beatngu13 (approved):

PR #560, opened 2019-12-08
Feature/refactorings

roesslerj:
- [] <s>the necessary tests are either created or updated.</s>

### DIFF
--- a/src/main/java/de/retest/recheck/ui/diff/RootElementDifference.java
+++ b/src/main/java/de/retest/recheck/ui/diff/RootElementDifference.java
@@ -155,4 +155,8 @@ public class RootElementDifference implements Difference {
 	public String getTitle() {
 		return title;
 	}
+
+	public RootElement getElement() {
+		return getExpectedDescriptor() != null ? getExpectedDescriptor() : getActualDescriptor();
+	}
 }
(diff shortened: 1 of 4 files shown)


src/main/java/de/retest/recheck/ui/diff/RootElementDifference.java
```diff
@@ -163,4 +155,8 @@ public int getCheckedUiComponentCount() {
 	public String getTitle() {
 		return title;
 	}
+
+	public RootElement getElement() {
```
beatngu13, line 175: I suggest that we either call this method `getDescriptor()` (then it would correspond to the other methods), or even better: refactor the term "descriptor" to "element" in this class.

beatngu13 (changes requested):

src/main/java/de/retest/recheck/ui/diff/RootElementDifference.java
roesslerj: Ok, renamed the other methods instead.

beatngu13 (approved):

beatngu13:
SonarCloud complains:

https://sonarcloud.io/project/issues?id=de.retest%3Arecheck&pullRequest=560&resolved=false&types=CODE_SMELL

Actually, why deprecate at all? Its internal API and easy to replace?

roesslerj:
> Its internal API and easy to replace?

You sure there are no other dependencies? Given past experiences, I'd rather be super careful and do this in two steps...

PR #574, opened 2019-12-11
Feature/fix metadata

diba1013:

- [x] ~you updated the changelog.~

## Description

> TL;DR Fix some Kryo (de)serialization issues and reversed parameters

### DIFF
--- a/src/main/java/de/retest/recheck/RecheckImpl.java
+++ b/src/main/java/de/retest/recheck/RecheckImpl.java
@@ -133,7 +133,7 @@ public class RecheckImpl implements Recheck, SutStateLoader {
 		final RecheckDifferenceFinder finder =
 				new RecheckDifferenceFinder( defaultFinder, currentStep, file.getPath() );
 
-		final ActionReplayResult actionReplayResult = finder.findDifferences( actual, expected );
+		final ActionReplayResult actionReplayResult = finder.findDifferences( expected, actual );
 		if ( actionReplayResult.hasDifferences() ) {
 			adapter.notifyAboutDifferences( actionReplayResult );
 		}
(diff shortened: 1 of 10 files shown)


src/main/java/de/retest/recheck/ui/diff/meta/MetadataDifferenceFinder.java
```diff
@@ -10,24 +10,24 @@
 
 public class MetadataDifferenceFinder {
 
-	public MetadataDifference findDifferences( final SutState expected, final SutState actual ) {
-		return findDifferences( expected.getMetadata(), actual.getMetadata() );
+	public MetadataDifference findDifferences( final SutState actual, final SutState expected ) {
```
roesslerj: Everywhere else, it is `expected, actual` ... I would keep that order to be consistent.

roesslerj (changes requested):

src/main/java/de/retest/recheck/ui/diff/meta/MetadataDifferenceFinder.java
diba1013: Well, that's embarrassing... 
`RecheckDifferenceFinder#findDifferences` is changed, therefore I assumed this was mixed, too.

Will revert that

roesslerj: Oh. We should fix `RecheckDifferenceFinder#findDifferences` as well...

diba1013: Done

roesslerj (approved):

PR #575, opened 2019-12-12
Fix defaults

beatngu13:
Just realized we haven't documented the properties from the default file. Will open a PR for this as well.

### DIFF
--- a/src/main/resources/default-recheck.ignore
+++ b/src/main/resources/default-recheck.ignore
@@ -19,7 +19,6 @@
 # https://docs.retest.de/recheck/usage/filter/
 
 # These are sensible defaults, delete if they don't apply:
-attribute=absolute-outline
 attribute=style
 matcher: type=input, attribute-regex: border-.*-color
 pixel-diff=5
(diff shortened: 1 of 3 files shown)


src/main/resources/default-recheck.ignore
```diff
@@ -1,27 +1,24 @@
...
-# matcher: xpath=HTML[1]/BODY[1]/DIV[1]
+# Or via absolute XPath:
+# matcher: xpath=html[1]/body[1]/div[1]
 
 # To ignore attributes globally, use:
 # attribute=class
 
-# To ignore attributes e.g. of specific elements: 
+# To ignore attributes e.g. of specific elements, use (excludes children): 
 # matcher: type=img, attribute: alt
```
diba1013: This is now also `=`
```suggestion
# matcher: type=img, attribute=alt
```

diba1013 (changes requested):

src/main/resources/default-recheck.ignore
beatngu13: I will never learn this. 😉

diba1013 (approved):

diba1013 (approved):

PR #596, opened 2020-01-10
OptionalUtil

beatngu13:

## Description

Rename `StreamUtil` to `OptionalUtil` as this is essentially `Optional#stream()` from Java 9, and add `or` helper also from Java 9.

For more details, please refer to the commit messages.

### DIFF
--- a/src/main/java/de/retest/recheck/util/OptionalUtil.java
+++ b/src/main/java/de/retest/recheck/util/OptionalUtil.java
@@ -1,6 +1,7 @@
 package de.retest.recheck.util;
 
 import java.util.Optional;
+import java.util.function.Supplier;
 import java.util.stream.Stream;
 
 public class OptionalUtil {
@@ -9,6 +10,10 @@ public class OptionalUtil {
 
 	public static <T> Stream<T> stream( final Optional<T> o ) {
 		return o.map( Stream::of ).orElseGet( Stream::empty );
+	}
+
+	public static <T> Optional<T> or​( final Optional<T> o, final Supplier<Optional<T>> s ) {
+		return o.isPresent() ? o : s.get();
 	}
 
 }
--- a/src/main/java/de/retest/recheck/util/OptionalUtil.java
+++ b/src/main/java/de/retest/recheck/util/OptionalUtil.java
@@ -3,11 +3,11 @@ package de.retest.recheck.util;
 import java.util.Optional;
 import java.util.stream.Stream;
 
-public class StreamUtil {
+public class OptionalUtil {
 
-	private StreamUtil() {}
+	private OptionalUtil() {}
 
-	public static <T> Stream<T> optionalToStream( final Optional<T> o ) {
+	public static <T> Stream<T> stream( final Optional<T> o ) {
 		return o.map( Stream::of ).orElseGet( Stream::empty );
 	}
 
--- a/src/main/java/de/retest/recheck/util/OptionalUtil.java
+++ b/src/main/java/de/retest/recheck/util/OptionalUtil.java
@@ -8,12 +8,15 @@ public class OptionalUtil {
 
 	private OptionalUtil() {}
 
-	public static <T> Stream<T> stream( final Optional<T> o ) {
-		return o.map( Stream::of ).orElseGet( Stream::empty );
+	@SuppressWarnings( "unchecked" )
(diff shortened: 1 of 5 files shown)


src/main/java/de/retest/recheck/util/OptionalUtil.java
```diff
@@ -0,0 +1,22 @@
+package de.retest.recheck.util;
+
+import java.util.Optional;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+public class OptionalUtil {
```
diba1013, line 7: If we only could use [`@ExtensionMethod`](https://projectlombok.org/features/experimental/ExtensionMethod)

src/main/java/de/retest/recheck/util/OptionalUtil.java
```diff
@@ -0,0 +1,22 @@
...
+
+	private OptionalUtil() {}
+
+	@SuppressWarnings( "unchecked" )
+	public static <T> Stream<T> stream( final Optional<? extends T> o ) {
+		return (Stream<T>) o.map( Stream::of ).orElseGet( Stream::empty );
+	}
```
diba1013, line 14: Do we want to use a real method and not a lambda?

```java
private static final Function<Optional<?>, Stream<?>> toStream = o -> o.map( Stream::of ).orElse( Stream.empty() );

public static <T> Function<? super Optional<T>, Stream<T> stream() {
    return /*Fancy cast here*/ toStream;
}
```

src/test/java/de/retest/recheck/util/OptionalUtilTest.java
```diff
@@ -0,0 +1,40 @@
...
+	@Test
+	void stream_should_be_empty_if_absent() throws Exception {
+		final Optional<Object> absent = Optional.empty();
+		assertThat( OptionalUtil.stream( absent ) ).isEmpty();
+	}
+
+	@Test
+	void or_should_return_given_optional_if_present() throws Exception {
+		final Optional<Object> opt = Optional.of( new Object() );
+		final Supplier<Optional<Object>> supplier = () -> Optional.empty();
```
diba1013: Do you want the same scheme as below with `absent` and `present`

diba1013 (changes requested):
Next PR "upgrade to java 11": The desperate cry ...

src/main/java/de/retest/recheck/util/OptionalUtil.java
beatngu13: Or [Kotlin](https://kotlinlang.org/docs/reference/extensions.html). 😉 But yes, I vote against experimental Lombok features in prod.

src/main/java/de/retest/recheck/util/OptionalUtil.java
beatngu13: Dunno, what is the benefit?

src/test/java/de/retest/recheck/util/OptionalUtilTest.java
beatngu13: Like:

```java
final Optional<Object> present = Optional.of( new Object() );
final Supplier<Optional<Object>> absent = () -> Optional.empty();
assertThat( OptionalUtil.or​( present, absent ) ).isEqualTo( present );
```

?

src/main/java/de/retest/recheck/util/OptionalUtil.java
diba1013: Maybe increased readability on site of the usage and&mdash;on a lower note&mdash;fewer objects created?

```java
list.stream() //
    .map( OptionalUtil::stream )

list.stream() //
    .map( OptionalUtil.stream() )
```

But same as the discussion about complex lambda vs method reference within a static method: No real preference.

src/test/java/de/retest/recheck/util/OptionalUtilTest.java
diba1013: Yes

src/main/java/de/retest/recheck/util/OptionalUtil.java
beatngu13: I think I personally prefer the current, method-ref-based style. But thanks for bringing attention to this.

diba1013 (approved):

PR #602, opened 2020-01-15
Feature/rehub convenience

modulo11:
Convenient methods to be called e.g. from CLI.


roesslerj (commented):
Are there any tests for this? Can't there be?

modulo11:
Unfortunately no. We'd need to mock the entire auth system which I haven't done yet.

roesslerj (approved):

PR #606, opened 2020-01-16
Feature/display warning

kema91:
We want to display the warnings for each attribute difference.

### DIFF
--- a/src/main/java/de/retest/recheck/printer/AttributeDifferencePrinter.java
+++ b/src/main/java/de/retest/recheck/printer/AttributeDifferencePrinter.java
@@ -1,10 +1,13 @@
 package de.retest.recheck.printer;
 
 import java.io.Serializable;
+import java.util.List;
+import java.util.stream.Collectors;
 
 import de.retest.recheck.ui.DefaultValueFinder;
 import de.retest.recheck.ui.descriptors.IdentifyingAttributes;
 import de.retest.recheck.ui.diff.AttributeDifference;
+import de.retest.recheck.ui.diff.ElementIdentificationWarning;
 import lombok.RequiredArgsConstructor;
 
 @RequiredArgsConstructor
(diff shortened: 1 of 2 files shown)


src/main/java/de/retest/recheck/printer/AttributeDifferencePrinter.java
```diff
@@ -58,4 +66,15 @@ private String printBothDifferences( final AttributeDifference difference ) {
 		final Serializable actual = difference.getActual();
 		return String.format( KEY_EXPECTED_ACTUAL_FORMAT, key, expected, actual );
 	}
+
+	private String printWarning( final AttributeDifference difference ) {
+		final Integer warnings = difference.getElementIdentificationWarnings().size();
```
modulo11: Any reason to use a boxed integer?

src/main/java/de/retest/recheck/printer/AttributeDifferencePrinter.java
```diff
@@ -58,4 +66,15 @@ private String printBothDifferences( final AttributeDifference difference ) {
 		final Serializable actual = difference.getActual();
 		return String.format( KEY_EXPECTED_ACTUAL_FORMAT, key, expected, actual );
 	}
+
+	private String printWarning( final AttributeDifference difference ) {
```
modulo11, line 70: Why are `fileName` and `line` only extracted when there is only one warning?

src/main/java/de/retest/recheck/printer/AttributeDifferencePrinter.java
```diff
@@ -14,11 +15,18 @@
 
 	private static final String KEY_EXPECTED_ACTUAL_FORMAT = "%s: expected=\"%s\", actual=\"%s\"";
 
+	private static final String WARNING_FILENAME_LINE_FORMAT = ", warning=\"%s:%d\"";
```
modulo11: Why is there a distinction between warning and warnings? What does warning mean for a user? Maybe there is a more expressive name (at least for displaying it).

diba1013:
Since @kema91 is currently on vacation, I took over this PR to have it merged in the upcoming release.

I settled for the following changes:

1. Display all warnings from a `AttributeDifference`. However, currently, this should not happen as only one attribute can be used as identifying.
2. Rename "warning" to "breaks". What this term means cannot be deeply explained within the CLI output as it would clutter the message. Instead we should take care that this is properly documented.

modulo11 (approved):

PR #616, opened 2020-01-24
Dislpay the actual filename (as is said in the message) not the description

roesslerj:


diba1013:
Changelog, as this affects review and recheck.cli?

src/main/java/de/retest/recheck/suite/flow/ApplyChangesToStatesFlow.java
```diff
@@ -73,7 +73,7 @@ private ApplyChangesToStatesFlow( final Persistence<SutState> persistence ) {
 				return Collections.emptyList();
 			}
 			goldenMasterProvider.saveGoldenMaster( file, newState );
-			return Collections.singletonList( actionChangeSet.getDescription() );
+			return Collections.singletonList( file.getName() );
```
diba1013, line 76: Do we want to transform this to a absolute path?

diba1013 (commented):
Changelog, as this affects review and recheck.cli?

src/main/java/de/retest/recheck/suite/flow/ApplyChangesToStatesFlow.java
roesslerj: I'd say we go with this for now and see how to further improve it later...

roesslerj:
CHANGELOG adapted

diba1013 (approved):

PR #617, opened 2020-01-24
Feature/not append recheck ignore

roesslerj:

## Description

Fix that if there are multiple recheck.ignore files, that eventually all are appended into one. Therefore I think this is a fix for a bug...

### DIFF
--- a/src/main/java/de/retest/recheck/ignore/PersistentFilter.java
+++ b/src/main/java/de/retest/recheck/ignore/PersistentFilter.java
@@ -22,6 +22,6 @@ public class PersistentFilter implements Filter {
 	private final Filter filter;
 
 	public static List<Filter> unwrap( final List<Filter> input ) {
-		return input.stream().map( f -> ((PersistentFilter) f).getFilter() ).collect( Collectors.toList() );
+		return input.stream().map( f -> f instanceof PersistentFilter ? ((PersistentFilter) f).getFilter() : f ).collect( Collectors.toList() );
 	}
 }
--- a/src/main/java/de/retest/recheck/ignore/PersistentFilter.java
+++ b/src/main/java/de/retest/recheck/ignore/PersistentFilter.java
@@ -5,12 +5,13 @@ import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import de.retest.recheck.ui.descriptors.Element;
+import de.retest.recheck.ui.diff.AttributeDifference;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.ToString;
-import lombok.experimental.Delegate;
 
-@ToString
+@ToString( onlyExplicitlyIncluded = true )
 @AllArgsConstructor
 public class PersistentFilter implements Filter {
 
@@ -18,15 +19,35 @@ public class PersistentFilter implements Filter {
 	private final Path path;
 
 	@ToString.Include
-	@Delegate
 	@Getter
+	// Delegate
 	private final Filter filter;
 
 	public static Stream<PersistentFilter> wrap( final Path path, final Stream<Filter> filter ) {
 		return filter.map( f -> new PersistentFilter( path, f ) );
 	}
 
-	public static List<Filter> unwrap( final List<Filter> input ) {
-		return input.stream().map( f -> f instanceof PersistentFilter ? ((PersistentFilter) f).getFilter() : f ).collect( Collectors.toList() );
+	public static List<Filter> unwrap( final List<? extends Filter> input ) {
+		return input.stream().map( f -> f instanceof PersistentFilter ? ((PersistentFilter) f).getFilter() : f )
+				.collect( Collectors.toList() );
 	}
+
+	@Override
+	public boolean matches( final Element element ) {
+		// Delegate
+		return filter.matches( element );
+	}
+
+	@Override
+	public boolean matches( final Element element, final String attributeKey ) {
+		// Delegate
+		return filter.matches( element, attributeKey );
+	}
+
+	@Override
+	public boolean matches( final Element element, final AttributeDifference attributeDifference ) {
+		// Delegate
+		return filter.matches( element, attributeDifference );
+	}
+
 }
--- a/src/main/java/de/retest/recheck/ignore/PersistentFilter.java
+++ b/src/main/java/de/retest/recheck/ignore/PersistentFilter.java
@@ -3,6 +3,7 @@ package de.retest.recheck.ignore;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import lombok.AllArgsConstructor;
 import lombok.Getter;
@@ -20,6 +21,10 @@ public class PersistentFilter implements Filter {
 	@Delegate
 	@Getter
 	private final Filter filter;
+
+	public static Stream<PersistentFilter> wrap( final Path path, final Stream<Filter> filter ) {
+		return filter.map( f -> new PersistentFilter( path, f ) );
+	}
 
 	public static List<Filter> unwrap( final List<Filter> input ) {
 		return input.stream().map( f -> f instanceof PersistentFilter ? ((PersistentFilter) f).getFilter() : f ).collect( Collectors.toList() );
--- a/src/main/java/de/retest/recheck/ignore/PersistentFilter.java
+++ b/src/main/java/de/retest/recheck/ignore/PersistentFilter.java
@@ -1,0 +1,22 @@
+package de.retest.recheck.ignore;
+
+import java.nio.file.Path;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.ToString;
+import lombok.experimental.Delegate;
+
+@ToString
+@AllArgsConstructor
+public class PersistentFilter implements Filter {
+
+	@Getter
+	private final Path path;
+
+	@ToString.Include
+	@Delegate
+	@Getter
+	private final Filter filter;
+
+}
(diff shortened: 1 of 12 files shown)


diba1013 (changes requested):
I'd like to have some tests written for this behavior to ensure that it works properly. From manually testing with review it seems to work, but just to make sure that we have some new tests that especially test the correct splitting and writing of the corresponding filters.

src/main/java/de/retest/recheck/review/ignore/io/Loader.java
```diff
@@ -14,6 +17,10 @@
 				.flatMap( OptionalUtil::stream );
 	}
 
+	default Stream<PersistentFilter> load( final Path path, final Stream<String> lines ) {
```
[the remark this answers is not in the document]
roesslerj: Ok. Moved it to `PersistentFilter` alongside `unwrap`.

src/main/java/de/retest/recheck/review/GlobalIgnoreApplier.java
```diff
@@ -58,7 +62,9 @@ public void unignoreElement( final Element element ) {
 	}
 
 	private void add( final Filter filter ) {
-		filtered.add( filter );
+		// TODO Receive target path from GUI
```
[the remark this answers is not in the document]
roesslerj: Difficult, because this is a pretty low-level task. I'd just leave it as is for now.

src/main/java/de/retest/recheck/review/workers/SaveFilterWorker.java
```diff
@@ -1,49 +1,52 @@
...
+				filters.collect( Collectors.groupingBy( PersistentFilter::getPath ) );
+
+		for ( final Map.Entry<Path, List<PersistentFilter>> entry : mapped.entrySet() ) {
+			final List<String> save = entry.getValue().stream() //
+					.map( f -> Loaders.filter().save( f.getFilter() ) ) //
+					.collect( Collectors.toList() );
+			final Path target = entry.getKey();
+			log.info( "Writing filters to file {}.", target );
+			createDirectories( target.getParent() );
+			write( target, save );
```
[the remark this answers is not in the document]
roesslerj: See SaveFilterWorkerIT#multiple_different_files_should_be_resolved_correctly.

src/main/java/de/retest/recheck/review/GlobalIgnoreApplier.java
```diff
@@ -58,7 +62,9 @@ public void unignoreElement( final Element element ) {
...
+		final Path ignorePath = locator.getProjectIgnoreFile().orElse( locator.getUserIgnoreFile() );
+		filtered.add( new PersistentFilter( ignorePath, filter ) );
```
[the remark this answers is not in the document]
roesslerj: Done in `GlobalIgnoreApplierTest`.

src/test/java/de/retest/recheck/review/workers/SaveFilterWorkerIT.java
```diff
@@ -23,70 +24,59 @@
...
+		filter.add( new PersistentFilter( file1, new FilterPreserveLine( "# C" ) ) );
+
+		final SaveFilterWorker cut = new SaveFilterWorker( createApplier( filter ) );
 
 		cut.save();
 
-		assertThat( base.resolve( ".retest" ) ).exists();
+		assertThat( file1.toFile() ).hasContent( "# A\n# C\n# D\n# C" );
+		assertThat( file2.toFile() ).hasContent( "# B" );
+		assertThat( file3.toFile() ).hasContent( "# E" );
```
diba1013: Same here, `Path`  should work here

src/main/java/de/retest/recheck/review/GlobalIgnoreApplier.java
```diff
@@ -31,6 +41,11 @@ public static GlobalIgnoreApplier create( final Counter counter, final Persistab
 		return new GlobalIgnoreApplier( counter, applier.getIgnores() );
 	}
 
+	// Just for testing
+	protected static GlobalIgnoreApplier create( final Counter counter, final RecheckIgnoreLocator locator ) {
```
diba1013: If we use this for testing, why not make this package private?

src/main/java/de/retest/recheck/ignore/PersistentFilter.java
```diff
@@ -0,0 +1,32 @@
...
+
+@ToString
+@AllArgsConstructor
+public class PersistentFilter implements Filter {
+
+	@Getter
+	private final Path path;
+
+	@ToString.Include
+	@Delegate
```
diba1013: We do not use experimental lombok annotations

src/main/java/de/retest/recheck/ignore/PersistentFilter.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.experimental.Delegate;
+
+@ToString
+@AllArgsConstructor
+public class PersistentFilter implements Filter {
+
+	@Getter
+	private final Path path;
+
+	@ToString.Include
```
diba1013, line 21: You would need to specify `@ToString(onlyExplicitlyIncluded = true)`  as described [here](https://projectlombok.org/features/ToString)

src/main/java/de/retest/recheck/review/workers/SaveFilterWorker.java
```diff
@@ -1,49 +1,52 @@
...
-		try ( final PrintStream writer = new PrintStream( FileUtils.openOutputStream( ignoreFile ) ) ) {
-			save.forEach( writer::println );
+				.filter( filter -> (filter instanceof PersistentFilter) ) //
+				.map( f -> (PersistentFilter) f );
+
+		final Map<Path, List<PersistentFilter>> mapped =
+				filters.collect( Collectors.groupingBy( PersistentFilter::getPath ) );
+
+		for ( final Map.Entry<Path, List<PersistentFilter>> entry : mapped.entrySet() ) {
+			final List<String> save = entry.getValue().stream() //
```
diba1013: Please rename this variable to `lines` or similar, so that it is more meaningful

src/test/java/de/retest/recheck/review/GlobalIgnoreApplierTest.java
```diff
@@ -1,46 +1,93 @@
...
+			@Override
+			public Path getUserIgnoreFile() {
+				return base.resolve( "userhome/recheck.ignore" );
+			}
+		};
+		final GlobalIgnoreApplier cut = GlobalIgnoreApplier.create( getInstance(), locateNonExistentUserHome );
+
+		cut.ignoreElement( createFake() );
+
+		assertThat( cut.persist().getIgnores().toString() ).contains( "project" );
```
diba1013: If you changed `GlobalIgnoreApplier` `List<Filter>` to `List<PersistFilter>` you could more elegantly check for the `Path`.

diba1013 (changes requested):

src/main/java/de/retest/recheck/ignore/PersistentFilter.java
roesslerj: Too bad...

src/test/java/de/retest/recheck/review/workers/SaveFilterWorkerIT.java
roesslerj: Why? Here I actually want an IT... so I check the file contents.

diba1013: Yes, I meant something like:
`assertThat(file1).hasContent("...");`

src/main/java/de/retest/recheck/ignore/PersistentFilter.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	public static List<Filter> unwrap( final List<? extends Filter> input ) {
+		return input.stream().map( f -> f instanceof PersistentFilter ? ((PersistentFilter) f).getFilter() : f )
+				.collect( Collectors.toList() );
+	}
+
+	@Override
+	public boolean matches( final Element element ) {
+		// Delegate
```
diba1013, line 37: Why the comments? It should be obvious. But in case it is not, why not just rename the `filter` to `delegate`

src/main/java/de/retest/recheck/review/workers/SaveFilterWorker.java
```diff
@@ -1,49 +1,52 @@
...
+		final Stream<PersistentFilter> filters = persist.getIgnores().stream() //
 				.map( this::extractCachedFilter ) //
-				.filter( filter -> !(filter instanceof JSFilterImpl) );
-		final Stream<String> save = Loaders.filter().save( filters );
-
-		final File ignoreFile = ignorePath.orElse( locator.getUserIgnoreFile() ).toFile();
-		try ( final PrintStream writer = new PrintStream( FileUtils.openOutputStream( ignoreFile ) ) ) {
-			save.forEach( writer::println );
+				.filter( filter -> (filter instanceof PersistentFilter) ) //
+				.map( f -> (PersistentFilter) f );
```
diba1013: You do not need to do this anymore. Since you now have `List<PersistentFilter`, which never can be a `CacheFilter`.

Instead I would sugggest to change ll 39-40.
```
final Map<Path, List<Filter>> mapped =
filters.collect( Collectors.groupingBy( PersistentFilter::getPath, Collectors.mapping( f-> extractCachedFilter(f.getFilter), Collectors.toList() ) ) );
```

diba1013: I just realized that you have removed the filtering of the JSFilter, please add a test for that.

diba1013 (changes requested):

src/main/java/de/retest/recheck/ignore/PersistentFilter.java
roesslerj: Because then the generated getter and setter methods are also renamed, and that I don't want.

diba1013 (approved):

PR #621, opened 2020-01-27
Fix metadata diff printer

beatngu13:

## Description

As discussed.

### DIFF
--- a/src/main/java/de/retest/recheck/printer/MetadataDifferencePrinter.java
+++ b/src/main/java/de/retest/recheck/printer/MetadataDifferencePrinter.java
@@ -1,33 +1,13 @@
 package de.retest.recheck.printer;
 
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Set;
 import java.util.stream.Collectors;
 
-import de.retest.recheck.meta.global.GitMetadataProvider;
-import de.retest.recheck.meta.global.MachineMetadataProvider;
-import de.retest.recheck.meta.global.OSMetadataProvider;
-import de.retest.recheck.meta.global.TimeMetadataProvider;
 import de.retest.recheck.ui.diff.meta.MetadataDifference;
 import de.retest.recheck.ui.diff.meta.MetadataElementDifference;
 
 public class MetadataDifferencePrinter implements Printer<MetadataDifference> {
 
 	private static final String KEY_EXPECTED_ACTUAL_FORMAT = "%s: expected=\"%s\", actual=\"%s\"";
-
-	// Only print a selected amount of differences that potentially can break a test
-	private static final Set<String> differencesToIgnore = new HashSet<>( Arrays.asList( //
-			GitMetadataProvider.VCS_NAME, //
-			GitMetadataProvider.BRANCH_NAME, //
-			GitMetadataProvider.COMMIT_NAME, //
-			MachineMetadataProvider.MACHINE_NAME, //
-			OSMetadataProvider.OS_ARCH, //
-			TimeMetadataProvider.DATE, //
-			TimeMetadataProvider.TIME, //
-			TimeMetadataProvider.ZONE, //
-			TimeMetadataProvider.OFFSET //
-	) );
 
 	@Override
 	public String toString( final MetadataDifference difference, final String indent ) {
@@ -37,13 +17,8 @@ public class MetadataDifferencePrinter implements Printer<MetadataDifference> {
 
 	private String printDifferences( final MetadataDifference difference, final String indent ) {
 		return difference.getDifferences().stream() //
-				.filter( this::shouldPrint ) //
 				.map( this::print ) //
 				.collect( Collectors.joining( "\n" + indent, "\n" + indent, "" ) );
-	}
-
-	private boolean shouldPrint( final MetadataElementDifference difference ) {
-		return !differencesToIgnore.contains( difference.getKey() );
 	}
 
 	private String print( final MetadataElementDifference difference ) {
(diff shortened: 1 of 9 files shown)


src/main/java/de/retest/recheck/printer/MetadataDifferencePrinter.java
```diff
@@ -33,19 +33,22 @@
 
 	@Override
 	public String toString( final MetadataDifference difference, final String indent ) {
-		final String prefix = "Metadata Differences:";
-		return indent + prefix + printDifferences( difference, indent + "\t" );
+		final String differences = printDifferences( difference, indent + "\t" );
+		return differences.isEmpty() ? "" : indent + "Metadata Differences:" + differences;
```
diba1013: Doesn't this then have a empty line in TestPrinter? Since currently printers are supposed to always "print" a line, since all lines are just stupidly concatenated...

Could you write a test for that?

diba1013 (changes requested):

src/main/java/de/retest/recheck/report/MetadifferenceFilter.java
```diff
@@ -0,0 +1,38 @@
...
+			TimeMetadataProvider.ZONE, //
+			TimeMetadataProvider.OFFSET //
+	) );
+
+	public MetadataDifference filter( final MetadataDifference metadataDifference ) {
+		return metadataDifference.getDifferences().stream() //
+				.filter( diff -> !differencesToIgnore.contains( diff.getKey() ) ) //
+				.collect( collectingAndThen( //
+						toSet(), //
+						diffs -> diffs.isEmpty() ? MetadataDifference.empty() : MetadataDifference.of( diffs ) ) );
```
diba1013: Do you mind, putting this into `MetadataDifference`? This would then essentially be something like `Optional#ofNullable`
```suggestion
						MetadataDifference::of;
```

src/test/java/de/retest/recheck/report/MetadifferenceFilterTest.java
```diff
@@ -0,0 +1,42 @@
...
+		cut = new MetadifferenceFilter();
+	}
+
+	@Test
+	void should_filter_ignored_differences() {
+		final MetadataElementDifference elementDiff =
+				new MetadataElementDifference( TimeMetadataProvider.TIME, "expected", "actual" );
+		final Set<MetadataElementDifference> diffs = new HashSet<>( Arrays.asList( elementDiff ) );
+		final MetadataDifference diff = MetadataDifference.of( diffs );
+		assertThat( cut.filter( diff ).getDifferences() ).isEmpty();
```
diba1013: `MetadataDifference` implements `Iterable`, thus you can do directly (I think):
```suggestion
		assertThat( cut.filter( diff ) ).isEmpty();
```

src/test/java/de/retest/recheck/report/MetadifferenceFilterTest.java
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	@Test
+	void should_not_filter_unignored_differences() {
+		final MetadataElementDifference elementDiff = new MetadataElementDifference( "key", "expected", "actual" );
+		final Set<MetadataElementDifference> diffs = new HashSet<>( Arrays.asList( elementDiff ) );
+		final MetadataDifference diff = MetadataDifference.of( diffs );
+		assertThat( cut.filter( diff ).getDifferences() ).containsExactly( elementDiff );
```
diba1013: Same here...
```suggestion
		assertThat( cut.filter( diff ) ).containsExactly( elementDiff );
```

diba1013 (changes requested):
Do we need an integration test that verifies that these printer problems are resolved? A basic idea would be:

1. Since the metadata of a `RecheckAdapter` can overwrite the `GlobalMetadataProvider` you could enforce changes via a custom `RecheckAdapter`. 
2. Adapt `RecheckImplIT` accordingly.

There are still some test failures.

src/main/java/de/retest/recheck/report/MetadifferenceFilter.java
beatngu13: Good idea!

beatngu13:
>Do we need an integration test that verifies that these printer problems are resolved?

Don't think it's worth the effort.

But as we now have `of` and `ofEmptyable`, the question is if we want to check for the set size to be > 0 and raise an exception if this is not the case.

src/main/java/de/retest/recheck/ui/diff/meta/MetadataDifference.java
```diff
@@ -23,23 +25,26 @@
...
-		final Set<MetadataElementDifference> unique =
-				new TreeSet<>( Comparator.comparing( MetadataElementDifference::getKey ) );
-		unique.addAll( differences );
-		return new MetadataDifference( new LinkedHashSet<>( unique ) );
+		return differences.isEmpty() ? empty() : differences.stream() //
+				.collect( collectingAndThen( //
+						// filter duplicate keys (keep first seen)
+						toMap( MetadataElementDifference::getKey, Function.identity(), ( first, second ) -> first ), //
+						// construct with filtered metadata
+						map -> new MetadataDifference( new HashSet<>( map.values() ) ) ) ); //
```
diba1013: I like streams and I did something similar in review, but either solution is not that readable. In theory this should never happen... . Thus I would suggest leaving this as is and discuss this separately (since the current approach has the nice benefit of the differences being sorted, if that is any argument considering).

diba1013 (changes requested):
Still test failures 💥 😄 and commit history.

diba1013 (approved):

PR #627, opened 2020-01-29
Implement cascading filter referencing

roesslerj:

Add ability to import other filters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Table of Contents
 ### New Features
 
 * The `ProjectLayout` may now retrieve the test source root for the current project in order to allow for test healing to be applied. This is required for custom `ProjectLayouts` to be implemented, in order for this feature to work. For more information, please refer to the documentation.
+* Use `import: $reference` to import another filter into an filter or an ignore file. Ignoring all positioning changes is now easily possible with `import: positioning.filter`. Whole cascades and hierarchies of filters can now be created this way. 
 
 ### Improvements
 
--- a/src/main/java/de/retest/recheck/review/ignore/ImportedExternalFilter.java
+++ b/src/main/java/de/retest/recheck/review/ignore/ImportedExternalFilter.java
@@ -1,0 +1,45 @@
+package de.retest.recheck.review.ignore;
+
+import de.retest.recheck.ignore.Filter;
+import de.retest.recheck.review.ignore.io.ImportExternalFilterLoader;
+import de.retest.recheck.ui.descriptors.Element;
+import de.retest.recheck.ui.diff.AttributeDifference;
+
+public class ImportedExternalFilter implements Filter {
+
+	private final String reference;
+	private final Filter loaded;
+
+	public ImportedExternalFilter( final String reference, final Filter loaded ) {
+		this.loaded = loaded;
+		this.reference = reference;
+	}
+
+	@Override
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ Table of Contents
 ### New Features
 
 * The `ProjectLayout` may now retrieve the test source root for the current project in order to allow for test healing to be applied. This is required for custom `ProjectLayouts` to be implemented, in order for this feature to work. For more information, please refer to the documentation.
+* Use `import: $reference` to import another filter into an filter or an ignore file. Ignoring all positioning changes is now easily possible with `import: positioning.filter`. Whole cascades and hierarchies of filters can now be created this way.
```
diba1013, line 38: Do we need to make this more future proof?

``` 
import: name=positioning(.filter)?
import: url=https://
import: url=file://
import: file=/Users/with/absolute/path
import: file=relative/path/to/project
```
* Of course, do not implement this with this PR, just change the parsing, if you think: "yes"

Otherwise please stick to the given syntax:

- `:` should delegate to another loader. You could argue that this is the case but it could possible ruin/make it more difficult to implement the rules above (if we deem necessary). 
- `=` should be responsible for loading. We could make this then default/shortcut for `import: filter=${name}`

diba1013 (commented):
This is just my remark, please still review.

CHANGELOG.md
roesslerj: I think the syntax should be defined from a users perspective. Whether an implementation delegates to another loader is an implementation detail that the user doesn't care and shouldn't need to care about and that can be changed if needed, i.e. it's not a strong necessity but merely "accidental" due to a well-thought design.

As for the proposed extension (name/url/file), I don't see the benefit. As you already show by the examples, `url` and `file` are somewhat redundant, as `url` is a superset of `file`. And I don't see what benefit `name` yields. So I oppose that suggestion.

githubert, line 38: Use _a filter_.

githubert (changes requested):
Apart from a minor typo, it looks good. One small question: I haven't looked into the internals, but can this  feature cause a recursion, or are filters which are loaded not further parsed?

CHANGELOG.md
roesslerj: Sorry ... I don't see it. Where is the typo?

roesslerj:
> Apart from a minor typo, it looks good. One small question: I haven't looked into the internals, but can this feature cause a recursion, or are filters which are loaded not further parsed?

Wow, that's actually a great remark. Will look into that.

githubert (approved):

CHANGELOG.md
githubert: Damn, I was too quick with approving :D. It would have been at:

> to import another filter into **an filter** or an ignore file

roesslerj: Sorry, I still don't see it. Isn't this already **an filter**?

diba1013:
```suggestion
* Use `import: $reference` to import a filter into another filter or an ignore file. Ignoring all positioning changes is now easily possible with `import: positioning.filter`. Whole cascades and hierarchies of filters can now be created this way. 
```

roesslerj: Now I see it. Thought the typo was in filter ... Thanks.

PR #633, opened 2020-01-31
Feature/reduce redundancy

roesslerj:
Just a refactoring

### DIFF
--- a/src/main/java/de/retest/recheck/review/ignore/MatcherFilter.java
+++ b/src/main/java/de/retest/recheck/review/ignore/MatcherFilter.java
@@ -70,19 +70,12 @@ public class MatcherFilter implements Filter {
 
 		@Override
 		protected Optional<Filter> load( final MatchResult regex ) {
-			String match = regex.group( 1 );
-			if ( !match.contains( "," ) ) {
-				return Loaders.elementMatcher().load( match ) //
-						.map( matcher -> new MatcherFilter( matcher ) );
-			}
-			final String remainder = match.substring( match.indexOf( ',' ) + 1 ).trim();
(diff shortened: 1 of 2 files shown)


src/main/java/de/retest/recheck/review/ignore/MatcherFilter.java
```diff
@@ -70,19 +70,8 @@ public String save( final Filter ignore ) {
...
-			final String remainder = match.substring( match.indexOf( ',' ) + 1 ).trim();
-			match = match.substring( 0, match.indexOf( ',' ) );
-			// TODO Either no optional as return or no exception below
-			final Filter chained = chainableFilter.load( remainder ). //
-					orElseThrow( () -> new IllegalArgumentException(
-							"Couldn't find a filter for the expression '" + remainder + "'." ) );
-			return Loaders.elementMatcher().load( match ) //
-					.map( matcher -> new AllMatchFilter( new MatcherFilter( matcher ), chained ) );
+			return ChainableFilterLoaderUtil.load( regex,
+					match -> new MatcherFilter( Loaders.elementMatcher().load( match ).get() ) );
```
beatngu13: You shouldn't use `get` on an `Optional` without checking if the value is present. `map` as above?

beatngu13 (changes requested):
You have a test failure: ﻿https://travis-ci.com/retest/recheck/builds/147002875#L6349

src/main/java/de/retest/recheck/review/ignore/MatcherFilter.java
roesslerj: Thanks

PR #634, opened 2020-01-31
Feature/add colordiff

roesslerj:

Adds a percentual color-diff filter, much like the pixel filter. This allows to ignore, e.g. minor color changes or changes where opacity value is added or missing. Documentation is in the making...

### DIFF
--- a/src/main/java/de/retest/recheck/review/ignore/ColorDiffFilter.java
+++ b/src/main/java/de/retest/recheck/review/ignore/ColorDiffFilter.java
@@ -59,16 +59,13 @@ public class ColorDiffFilter implements Filter {
 	 * 100% change in the perceived color.
 	 */
 	static double calculateColorDistance( final Color expected, final Color actual ) {
-		final double r1 = expected.getRed();
-		final double r2 = actual.getRed();
-		final double g1 = expected.getGreen();
-		final double g2 = actual.getGreen();
-		final double b1 = expected.getBlue();
-		final double b2 = actual.getBlue();
-		final double a1 = expected.getAlpha();
-		final double a2 = actual.getAlpha();
+		final double redDiff = abs( expected.getRed() - actual.getRed() );
+		final double greenDiff = abs( expected.getGreen() - actual.getGreen() );
+		final double blueDiff = abs( expected.getBlue() - actual.getBlue() );
+		final double alphaDiff = abs( expected.getAlpha() - actual.getAlpha() );
 
-		final double distance = min( abs( r1 - r2 ) + abs( g1 - g2 ) + abs( b1 - b2 ) + abs( a1 - a2 ), 255 );
+		// min so we don't exceed max_distance
+		final double distance = min( redDiff + greenDiff + blueDiff + alphaDiff, 255 );
 		return distance / MAX_DISTANCE;
 	}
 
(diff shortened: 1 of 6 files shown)


src/main/java/de/retest/recheck/review/ignore/ColorDiffFilter.java
```diff
@@ -0,0 +1,125 @@
...
+import de.retest.recheck.ui.diff.AttributeDifference;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+@Getter
+@Slf4j
+public class ColorDiffFilter implements Filter {
+
+	private static final Pattern RGB = Pattern.compile( "rgb\\((\\d+),\\s?(\\d+),\\s?(\\d+)(,\\s?(\\d+))?\\)" );
+	private static final double MAX_DISTANCE = 255;
```
modulo11: `double` declared, but `int` given.

src/main/java/de/retest/recheck/review/ignore/ColorDiffFilter.java
```diff
@@ -0,0 +1,125 @@
...
+		return distance < colorDiff;
+	}
+
+	/**
+	 * Return the color distance as a value in the range of [0.0, 1.0]. Note this is about perception: any one of the
+	 * rgba values is reviewed in isolation, as a change from e.g. white to red or green to yellow already constitutes a
+	 * 100% change in the perceived color.
+	 */
+	static double calculateColorDistance( final Color expected, final Color actual ) {
+		final double r1 = expected.getRed();
```
modulo11: It may not look nice in L71, but can you use some more descriptive name schema like `redActual`/`redExpected` and so on?

src/main/java/de/retest/recheck/review/ignore/ColorDiffFilter.java
```diff
@@ -0,0 +1,125 @@
...
+		final String color = (String) input;
+		if ( color.startsWith( "#" ) ) {
+			return Color.decode( color );
+		}
+		final Matcher matcher = RGB.matcher( color );
+		if ( !matcher.find() ) {
+			return null;
+		}
+		try {
+			final String r = matcher.group( 1 );
```
modulo11: Same as L62.

modulo11 (changes requested):

src/main/java/de/retest/recheck/review/ignore/ColorDiffFilter.java
roesslerj: Why is that a problem? We need this to be a double later in the calculation.

src/main/java/de/retest/recheck/review/ignore/ColorDiffFilter.java
roesslerj: Done

src/main/java/de/retest/recheck/review/ignore/ColorDiffFilter.java
roesslerj: Done

src/main/java/de/retest/recheck/review/ignore/ColorDiffFilter.java
modulo11: No need for a conversion here. You can just initialize it with a double:
`private static final double MAX_DISTANCE = 255.0;`

modulo11 (changes requested):

modulo11 (approved):

PR #635, opened 2020-01-31
Add a ValueFilter that allows to ignore e.g. specific value pattern

roesslerj:

Can now specify that a specific value-pattern should be filtered, either globally, for specific elements or specific attributes. This allows to ignore a date, but ensure that it is actually a valid date.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Table of Contents
 ### New Features
 
 * You can now specify that e.g. the pixel-diff filter should only apply to a specific subtree of the given elements, or to only a specific attribute. For more details, see the [docs](https://docs.retest.de).
+* You can now specify that a specific value-pattern should be filtered, either globally, for specific elements or specific attributes. This allows to ignore a date, but ensure that it is actually a valid date.
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ Table of Contents
 ### New Features
 
 * You can now specify that e.g. the pixel-diff filter should only apply to a specific subtree of the given elements, or to only a specific attribute. For more details, see the [docs](https://docs.retest.de).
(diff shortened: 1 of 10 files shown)


src/main/java/de/retest/recheck/review/ignore/ValuePatternFilter.java
```diff
@@ -0,0 +1,67 @@
...
+	public boolean matches( final Element element, final AttributeDifference attributeDifference ) {
+		final Serializable actual = attributeDifference.getActual();
+
+		if ( actual == null || !(actual instanceof String) ) {
+			return false;
+		}
+
+		final Matcher matcher = pattern.matcher( (String) actual );
+
+		if ( matcher.matches() ) {
```
modulo11: Should be safe to return `matcher.matches()`

modulo11 (changes requested):

src/main/java/de/retest/recheck/review/ignore/ValuePatternFilter.java
roesslerj: Thanks, Done

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ Table of Contents
 ### New Features
 
 * You can now specify that e.g. the pixel-diff filter should only apply to a specific subtree of the given elements, or to only a specific attribute. For more details, see the [docs](https://docs.retest.de).
+* You can now specify that a specific value-pattern should be filtered, either globally, for specific elements or specific attributes. This allows to ignore a date, but ensure that it is actually a valid date.
```
diba1013: Could you add such cases to the recheck.ignore within the test resource root (`SaveAndLoadFilterWorker`), as to identify these use cases (as I understand from this changelog):
```
value-regex=.*
attribute=foo, value-regex=.*
element: id=foo, value-regex=.*
element: id=foo, attribute=bar, value-regex=.*
```

diba1013 (dismissed):

CHANGELOG.md
roesslerj: Done

modulo11 (approved):

PR #637, opened 2020-02-05
Feature/fix class matcher

roesslerj:

Fixed a bug where, when filtering for classes, the given classes had to contain _all_ of the elements classes, not the other way around.

### DIFF
--- a/src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
+++ b/src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
@@ -31,7 +31,7 @@ public class ElementClassMatcher implements Matcher<Element> {
 		if ( classValues.isEmpty() ) {
 			return false;
 		}
-		return this.classValues.containsAll( classValues );
+		return classValues.containsAll( this.classValues );
 	}
 
 	@Override
--- a/src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
+++ b/src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
@@ -15,28 +15,28 @@ public class ElementClassMatcher implements Matcher<Element> {
 
 	public static final String CLASS_KEY = "class";
 
-	private final List<String> classValues;
+	private final List<String> classes;
 
 	public ElementClassMatcher( final Element element ) {
 		this( toClassValuesList( toClassValueString( element ) ) );
 	}
 
 	private ElementClassMatcher( final List<String> classValues ) {
-		this.classValues = classValues;
+		classes = classValues;
 	}
 
 	@Override
 	public boolean test( final Element element ) {
-		final List<String> classValues = toClassValuesList( toClassValueString( element ) );
-		if ( classValues.isEmpty() ) {
+		final List<String> elementClasses = toClassValuesList( toClassValueString( element ) );
+		if ( classes.isEmpty() ) {
 			return false;
 		}
-		return classValues.containsAll( this.classValues );
+		return elementClasses.containsAll( classes );
 	}
 
 	@Override
 	public String toString() {
-		final String classValueString = classValues.stream().collect( Collectors.joining( " " ) );
+		final String classValueString = classes.stream().collect( Collectors.joining( " " ) );
 		return String.format( ElementClassMatcherLoader.FORMAT, classValueString );
 	}
 
--- a/src/test/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcherTest.java
+++ b/src/test/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcherTest.java
@@ -51,17 +51,16 @@ class ElementClassMatcherTest {
 
 	@Test
 	void should_match_multiple_equal_class_values() throws Exception {
-		final Element oneTwoOnone = mockElementWithClassValue( "one two" );
-
-		final ElementClassMatcher cut = new ElementClassMatcher( oneTwoOnone );
-
-		final Element one = mockElementWithClassValue( "oneone" );
-		final Element two = mockElementWithClassValue( "two" );
 		final Element oneTwo = mockElementWithClassValue( "one two" );
-		final Element twoOne = mockElementWithClassValue( "boo two one" );
 
-		assertThat( cut ).accepts( oneTwo, twoOne );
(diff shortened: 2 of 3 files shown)


src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
```diff
@@ -15,28 +15,28 @@
...
 	public static final String CLASS_KEY = "class";
 
-	private final List<String> classValues;
+	private final List<String> classes;
 
 	public ElementClassMatcher( final Element element ) {
 		this( toClassValuesList( toClassValueString( element ) ) );
 	}
 
 	private ElementClassMatcher( final List<String> classValues ) {
```
beatngu13, line 24: Please also rename the remaining "classValues" to "classes".

src/test/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcherTest.java
```diff
@@ -51,16 +51,17 @@ void should_not_match_any_class_value_if_class_key_is_absent() throws Exception
 
 	@Test
 	void should_match_multiple_equal_class_values() throws Exception {
-		final Element oneTwoOnone = mockElementWithClassValue( "one two onone" );
+		final Element oneTwoOnone = mockElementWithClassValue( "one two" );
```
beatngu13: Now this would be `oneTwo` (name clash below).

src/test/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcherTest.java
```diff
@@ -51,16 +51,17 @@ void should_not_match_any_class_value_if_class_key_is_absent() throws Exception
...
 
 		final ElementClassMatcher cut = new ElementClassMatcher( oneTwoOnone );
 
-		final Element one = mockElementWithClassValue( "one" );
+		final Element one = mockElementWithClassValue( "oneone" );
 		final Element two = mockElementWithClassValue( "two" );
 		final Element oneTwo = mockElementWithClassValue( "one two" );
-		final Element twoOne = mockElementWithClassValue( "two one" );
+		final Element twoOne = mockElementWithClassValue( "boo two one" );
```
beatngu13:
```suggestion
		final Element booTwoOne = mockElementWithClassValue( "boo two one" );
```

src/test/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcherTest.java
```diff
@@ -51,16 +51,17 @@ void should_not_match_any_class_value_if_class_key_is_absent() throws Exception
...
 
-		assertThat( cut ).accepts( one, two, oneTwo, twoOne );
+		assertThat( cut ).accepts( oneTwo, twoOne );
+		assertThat( cut ).rejects( one, two );
```
beatngu13: To recap, if the user specifies (as above):

```
matcher: class=one two
```

It should filter elements with classes `one two` and `boo two one`, but not `oneone` and `two`? That is, elements must at least contain all classes from the filter?

src/test/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcherTest.java
```diff
@@ -51,16 +51,17 @@ void should_not_match_any_class_value_if_class_key_is_absent() throws Exception
 
 	@Test
 	void should_match_multiple_equal_class_values() throws Exception {
-		final Element oneTwoOnone = mockElementWithClassValue( "one two onone" );
+		final Element oneTwoOnone = mockElementWithClassValue( "one two" );
 
 		final ElementClassMatcher cut = new ElementClassMatcher( oneTwoOnone );
 
-		final Element one = mockElementWithClassValue( "one" );
+		final Element one = mockElementWithClassValue( "oneone" );
```
beatngu13:
```suggestion
		final Element oneOne = mockElementWithClassValue( "oneone" );
```

beatngu13 (changes requested):

src/test/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcherTest.java
roesslerj: That's IMHO what I would intuitively expect. If I want to use different classes, I would create different rules.

src/main/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcher.java
roesslerj: Why is it important to forcibly shadow the member variable? That was IMHO the reason for this bug...

src/test/java/de/retest/recheck/review/ignore/matcher/ElementClassMatcherTest.java
diba1013: We discussed this when implementing, but I do not remember our decision.. 🤔 

I took some time to think about this, but I now agree with @roesslerj.

Consider the following example:
```
<div class="one two">
```

We currently have this the other way round, an element must specify a subset of the filter, but cannot contain a superset. 

this will never match against:
```
matcher: class=one
```
In order to match, the filter must specify all classes:
```
matcher: class=one two
```


The test does it the other way round. It assumes someone writes the filter:
```
matcher: class=one two onone
```
and checks against the following elements:
```
<div class="one">
<div class="two">
<div class="one two">
<div class="two one">
```

---

Coming from Selenium and/or css selectors, one would assume:

1. `By.className("one")` matches
2. `By.className("two")` matches
3. By.className("one two") Selenium cannot handle multiple classes... or I am doing something wrong...
4. `.one` matches
5. `.two` matches
6. `.one.two` matches
7. `.two.one` matches

Using this for our filters:
1. `class=one` matches
2. `class=two` matches
3. `class=one two` matches
4. `class=two one` matches
5. `class=one two onone` does not match

roesslerj: So ... do you agree or is there still something to do?

roesslerj: If there's nothing todo, please approve somebody...

beatngu13 (approved):

PR #641, opened 2020-02-06
Feature/Metadata Difference user info

lemoncurry:

## Description

> When printing differences, it is quite miss-leading if there are metadata differences but no other differences. It should be noted that these differences have no effect on the report.

### DIFF
--- a/src/main/java/de/retest/recheck/printer/MetadataDifferencePrinter.java
+++ b/src/main/java/de/retest/recheck/printer/MetadataDifferencePrinter.java
@@ -12,7 +12,9 @@ public class MetadataDifferencePrinter implements Printer<MetadataDifference> {
 	@Override
 	public String toString( final MetadataDifference difference, final String indent ) {
 		final String prefix = "Metadata Differences:";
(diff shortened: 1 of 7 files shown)


src/main/java/de/retest/recheck/printer/MetadataDifferencePrinter.java
```diff
@@ -12,7 +12,9 @@
 	@Override
 	public String toString( final MetadataDifference difference, final String indent ) {
 		final String prefix = "Metadata Differences:";
-		return indent + prefix + printDifferences( difference, indent + "\t" );
+		final String note = "\n\t  Please note that Metadata Differences do not affect the result and are not included"
```
beatngu13:
```suggestion
		final String note = "\n\t  Please note that metadata differences do not affect the result and are not included"
```

Or just "these differences" since we already say "Metadata Differences" above (which I also think should be "Metadata differences).

beatngu13 (changes requested):

beatngu13 (approved):

PR #644, opened 2020-02-06
Feature/Remove obsolete dependencies

lemoncurry:
- [ ] <s>the necessary tests are either created or updated.</s>

## Description

> Remove unused dependency 

- remove unnecessary dependency org.apache.commons:commons-exec
- reorder dependencies in accordnace with scope
- add dependency org.eclipse.persistence:org.eclipse.persistence.core:2.5.2 as direct dependency.

### Additional Context
[Maven dependency management recommendations](http://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html)
RET-1963


pom.xml
```diff
@@ -105,26 +105,18 @@
...
-
 		<dependency>
 			<groupId>com.google.guava</groupId>
 			<artifactId>guava</artifactId>
 			<version>28.2-jre</version>
 		</dependency>
 
+		<dependency>
+			<groupId>org.eclipse.persistence</groupId>
+			<artifactId>org.eclipse.persistence.core</artifactId>
```
roesslerj, line 116: Why do we need an additional dependency?

roesslerj (approved):

pom.xml
lemoncurry: According to `mvn dependency:analyze`:

```
[WARNING] Used undeclared dependencies found:
[WARNING]    org.eclipse.persistence:org.eclipse.persistence.core:jar:2.5.2:compile
```

The official recommendation as described in my PR says that
>  it is a good practice to explicitly specify the dependencies you are directly using in your own source code

[Link with official recommendations](http://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html)

This is why I suggested to add this dependency as direct dependency instead of having it as an indirect dependency, although Recheck uses it directly.

`mvn dependency:tree`:

```
[INFO] +- org.eclipse.persistence:org.eclipse.persistence.moxy:jar:2.5.2:compile
[INFO] |  +- org.eclipse.persistence:org.eclipse.persistence.core:jar:2.5.2:compile
[INFO] |  |  \- org.eclipse.persistence:org.eclipse.persistence.asm:jar:2.5.2:compile
[INFO] |  \- org.eclipse.persistence:org.eclipse.persistence.antlr:jar:2.5.2:compile
```

beatngu13: But we are not using it directly, it is only used by a dependency of us?

Anyways, sorry, I didn't see this discussion: https://github.com/retest/recheck/pull/651

lemoncurry: > But we are not using it directly, it is only used by a dependency of us?

`mvn dependency:analyze` should check for dependencies that are needed directly, according to [this](https://stackoverflow.com/a/4567730) answer, so it appears to be used directly e.g. in class `src/main/java/de/retest/recheck/persistence/xml/util/SessionLogDelegate.java`

lemoncurry: > Anyways, sorry, I didn't see this discussion: #651

No problem it was just a suggestion.

beatngu13: You are right, `SessionLogEntry` uses classes from there. Thanks for pointing this out, I will revert my PR.

PR #651, opened 2020-02-10
Remove unused dependency

beatngu13:
@roesslerj seems to be unused, also not required by Moxy. Do you know why added this in the past?


roesslerj (approved):

lemoncurry:
@beatngu13 see my comment for the reasoning [here](https://github.com/retest/recheck/pull/644#discussion_r376973872)

PR #665, opened 2020-02-25
Feature/improve to string

roesslerj:
Should improving the toString methods be mentioned in the Changelog?

### DIFF
--- a/src/main/java/de/retest/recheck/ui/descriptors/IdentifyingAttributes.java
+++ b/src/main/java/de/retest/recheck/ui/descriptors/IdentifyingAttributes.java
@@ -184,7 +184,6 @@ public class IdentifyingAttributes implements Serializable, Comparable<Identifyi
 	@Override
 	public String toString() {
 		final String type = getType();
-		final String text = get( "text" );
 		if ( type == null ) {
 			return "";
 		}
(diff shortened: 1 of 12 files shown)


src/main/java/de/retest/recheck/ui/descriptors/IdentifyingAttributes.java
```diff
@@ -184,18 +184,26 @@ public double match( final IdentifyingAttributes other ) {
...
+		final String text = get( "text" );
 		if ( text != null ) {
-			result += " [" + text + "]";
+			return result + " [" + text + "]";
 		}
-		return result;
+		final String id = get( "id" );
+		if ( id != null ) {
+			return result + " [" + id + "]";
+		}
```
beatngu13, line 201: Isn't `id` web-specific? At least I can only find `name` in the `identifyingAttributes` key list above.

beatngu13 (commented):
Regarding the changelog: I don't think it's mandatory in this case, but feel free to update it if you want to.

src/main/java/de/retest/recheck/ui/descriptors/IdentifyingAttributes.java
roesslerj: It is indeed, but by using the `retestId` in the `Element`s `toString` method, we address this. So this should be now almost irrelevant...

beatngu13 (approved):

PR #667, opened 2020-02-28
Feature/adapt element diff printer

roesslerj:

This is essentially a fix for another PR, changing the `toString` methods.

### DIFF
--- a/src/main/java/de/retest/recheck/printer/ElementDifferencePrinter.java
+++ b/src/main/java/de/retest/recheck/printer/ElementDifferencePrinter.java
@@ -21,7 +21,7 @@ public class ElementDifferencePrinter implements Printer<ElementDifference> {
 
 	private String createDescription( final ElementDifference difference ) {
 		final IdentifyingAttributes attributes = difference.getIdentifyingAttributes();
-		return attributes + " at '" + attributes.getPath() + "':";
+		return attributes.getType() + " (" + difference.getElement() + ") at '" + attributes.getPath() + "':";
 	}
 
 	private String createDifferences( final ElementDifference difference, final String indent ) {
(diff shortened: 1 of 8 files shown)


src/main/java/de/retest/recheck/printer/ElementDifferencePrinter.java
```diff
@@ -21,7 +21,7 @@ public String toString( final ElementDifference difference, final String indent
 
 	private String createDescription( final ElementDifference difference ) {
 		final IdentifyingAttributes attributes = difference.getIdentifyingAttributes();
-		return attributes + " at '" + attributes.getPath() + "':";
+		return difference.getElement() + " at '" + attributes.getPath() + "':";
```
diba1013: It is definitely an improvement to have no output `h1 [Header]`, but I miss the type of the element.

I tested this and think, that the type should still be displayed,, since the id may not always reflect this and it is very difficult to identify the element (since you have to parse the path, which can be very long). When analyzing the report, I often find myself ignoring the path and paying attention to the tag name (e.g. looking for a button) to identify an element of importance. With this change however, I always need to look at the path in order for the type to be known.

I would therefore suggest to change the `Element#toString` to `[type] ([retestid])`
```
bar (title) at 'foo[1]/bar[1]':
```

diba1013 (changes requested):

src/main/java/de/retest/recheck/printer/ElementDifferencePrinter.java
roesslerj: Ok, I think we can try having that. But that would be a separate PR, right?

roesslerj: After having thought about it, I would put it into the ElementDifferencePrinter. Let me update...

diba1013 (approved):

PR #668, opened 2020-02-28
Feature/name in retest

roesslerj:

I suggested the retestId is a better indicator of the element than the XPath ... time to correct the implementation to make this actually true :-|

As the CHANGELOG says: the ID is often generated, the text is not, so also prefer text over ID.

### DIFF
--- a/src/main/java/de/retest/recheck/ui/descriptors/idproviders/AbstractFirstNonNullRetestIdProvider.java
+++ b/src/main/java/de/retest/recheck/ui/descriptors/idproviders/AbstractFirstNonNullRetestIdProvider.java
@@ -19,8 +19,9 @@ abstract class AbstractFirstNonNullRetestIdProvider implements RetestIdProvider 
 		}
 		final String htmlId = normalizeAndCut( identifyingAttributes.get( "id" ) );
 		final String text = normalizeAndCut( identifyingAttributes.get( "text" ) );
+		final String name = normalizeAndCut( identifyingAttributes.get( "name" ) );
 		final String type = normalizeAndCut( cutTypeQualifier( identifyingAttributes.get( "type" ) ) );
-		final String id = returnFirstNonBlank( htmlId, text, type, randomUUID().toString() );
+		final String id = returnFirstNonBlank( htmlId, text, name, type, randomUUID().toString() );
 		return makeUnique( id );
 	}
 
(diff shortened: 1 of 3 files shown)


src/main/java/de/retest/recheck/ui/descriptors/idproviders/AbstractFirstNonNullRetestIdProvider.java
```diff
@@ -17,10 +17,11 @@ public String getRetestId( final IdentifyingAttributes identifyingAttributes ) {
 		if ( identifyingAttributes == null ) {
 			throw new NullPointerException( "Identifying attributes must not be null." );
 		}
-		final String htmlId = normalizeAndCut( identifyingAttributes.get( "id" ) );
 		final String text = normalizeAndCut( identifyingAttributes.get( "text" ) );
+		final String htmlId = normalizeAndCut( identifyingAttributes.get( "id" ) );
+		final String name = normalizeAndCut( identifyingAttributes.get( "name" ) );
 		final String type = normalizeAndCut( cutTypeQualifier( identifyingAttributes.get( "type" ) ) );
-		final String id = returnFirstNonBlank( htmlId, text, type, randomUUID().toString() );
+		final String id = returnFirstNonBlank( text, htmlId, name, type, randomUUID().toString() );
```
beatngu13, line 24: Ah, this makes sense! Text is indeed often more helpful than an (internal) ID.

However, we can easily have a test for this I guess?

beatngu13 (changes requested):

src/main/java/de/retest/recheck/ui/descriptors/idproviders/AbstractFirstNonNullRetestIdProvider.java
diba1013: Just to clarify, the text is for example the text of the paragraph?

For example, the paragraph containing my first sentence would be as id `just_to_clarify` (We have no tests for this, so I had to manually put this together).

However, this attribute does not seem to gather the name of a button (since the name may be wrapped in a `span` or complex inner html). Thus it is very difficult to get the actual text which is displayed in the button.

Should we therefore consider the `name` even stronger than the `id`? Since buttons, input-groups (e.g. radio button groups) may reference a name.

I am not sure what the best approach is here, so we need to test this...

roesslerj: ID is supposed to be unique, whereas names don't need to be. As for now I would prefer IDs over name, until we have a good counter example. If you already have such a counter example (real-world, not theoretical), I would be very interested to see that.

beatngu13 (approved):

PR #674, opened 2020-03-09
Feature/simplify mergify

diba1013:

- [X] the necessary tests are either created or updated.
- [X] all status checks (Travis CI, SonarCloud, etc.) pass.
- [X] ~you updated the changelog.~

## Description

> Simplify mergify config to include branch protection

While browsing the mergify documentation for branch protection, I stumbled upon some other simplifications:

1. Remove rules that are handled by branch protections status checks. Theoretically, we should be able to get rid of the approved-by condition, but as we require this a hard condition, I would leave that just in case.
2. Simplify equality for `value == 0` as the documentation (example rules) suggest that `value <= 0` only needs to be done for a set.
3. Replace the single quotes with double quotes&mdash;which was done by the automatic code formatter in VSCode. I would argue that these can be even removed, if desired.

### State of this PR

I have tested this (while not extensively) with my docs-test repository and mergify seems to respect the branch protection rules. However, I would like to go over those rules before merging this PR just to make sure.

Secondly, I would like to deploy this configuration on all repositories. Sadly, to my knowledge, there is no global configuration for all repositories.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -2,10 +2,6 @@ pull_request_rules:
   - name: 'rebase unreviewed non-release PRs'
     conditions:
       - head~=^(?!(release|hotfix)).*$
-      - status-success=Travis CI - Branch
-      - status-success=Travis CI - Pull Request
-      - status-success=SonarCloud Code Analysis
-      - status-success=verification/cla-signed
       - "#approved-reviews-by<=0"
       - "#changes-requested-reviews-by<=0"
     actions:
@@ -14,10 +10,6 @@ pull_request_rules:
   - name: 'merge non-release PRs with strict rebase'
     conditions:
       - head~=^(?!(release|hotfix)).*$
-      - status-success=Travis CI - Branch
-      - status-success=Travis CI - Pull Request
-      - status-success=SonarCloud Code Analysis
-      - status-success=verification/cla-signed
       - "#approved-reviews-by>=1"
       - "#changes-requested-reviews-by<=0"
     actions:
@@ -29,10 +21,6 @@ pull_request_rules:
   - name: 'merge release PRs with strict merge'
     conditions:
       - head~=^(release|hotfix).*$
-      - status-success=Travis CI - Branch
-      - status-success=Travis CI - Pull Request
-      - status-success=SonarCloud Code Analysis
-      - status-success=verification/cla-signed
       - "#approved-reviews-by>=1"
       - "#changes-requested-reviews-by<=0"
     actions:
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,5 +1,5 @@
 pull_request_rules:
-  - name: 'rebase unreviewed non-release PRs'
+  - name: "rebase unreviewed non-release PRs"
     conditions:
       - head~=^(?!(release|hotfix)).*$
       - "#approved-reviews-by=0"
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -1,46 +1,34 @@
...
+  - name: "rebase unreviewed non-release PRs"
     conditions:
       - head~=^(?!(release|hotfix)).*$
-      - status-success=Travis CI - Branch
-      - status-success=Travis CI - Pull Request
-      - status-success=SonarCloud Code Analysis
-      - status-success=verification/cla-signed
-      - "#approved-reviews-by<=0"
-      - "#changes-requested-reviews-by<=0"
+      - "#approved-reviews-by=0"
```
beatngu13, line 5: Could you point me to the docs where it says this is only needed for sets? Because according the [attributes section](https://doc.mergify.io/conditions.html#attributes) `approved-reviews-by` is an array of strings, and according to the [operators section](https://doc.mergify.io/conditions.html#operators) `=` checks for strict equality among the array values.

.mergify.yml
```diff
@@ -1,46 +1,34 @@
 pull_request_rules:
-  - name: 'rebase unreviewed non-release PRs'
+  - name: "rebase unreviewed non-release PRs"
```
beatngu13, line 2: Be careful when it comes to strings in YAML, especially when you say we could remove them entirely:

https://twitter.com/rogpeppe/status/1235958742828699649

😉

When it comes to single vs. double quotes: Probably makes sense to pick one variant and then escape accordingly.

.mergify.yml
diba1013: https://doc.mergify.io/conditions.html#operators

> This operator checks for the value to be lesser then or equal to the provided value. It’s usually used to compare against the length of an array using the # prefix.

Hmm, maybe I misinterpreted this wrongly this monday morning...

But regardless, the examples always check `#{$array}` with `=`
https://doc.mergify.io/examples.html#require-all-requested-reviews-to-be-approved

.mergify.yml
diba1013: Ok, sure, keeping them does not hurt. But since we use double quotes below, I would stick to them `"`&mdash;besides, I would need to reconfigure my formatter in the other case.

.mergify.yml
```diff
@@ -1,46 +1,34 @@
...
-      - status-success=verification/cla-signed
-      - "#approved-reviews-by<=0"
-      - "#changes-requested-reviews-by<=0"
+      - "#approved-reviews-by=0"
+      - "#changes-requested-reviews-by=0"
     actions:
       rebase: {}
 
-  - name: 'merge non-release PRs with strict rebase'
+  - name: "merge non-release PRs with strict rebase"
```
beatngu13, line 10: Test

beatngu13 (changes requested):

.mergify.yml
```diff
@@ -1,46 +1,34 @@
 pull_request_rules:
-  - name: 'rebase unreviewed non-release PRs'
+  - name: "rebase unreviewed non-release PRs"
     conditions:
       - head~=^(?!(release|hotfix)).*$
```
beatngu13: As discussed, only try to convert this to a (double-quoted) string to. Then all conditions are specified in the same manner.

beatngu13 (changes requested):

beatngu13 (approved):

diba1013:
This PR has to be manually merged.

PR #676, opened 2020-03-10
Fix/import filter

diba1013:

## Description

> TL;DR Fix importing filters not being able to load

### Problem

This should fix importing filters via:

```properties
# custom.filter
import: positioning.filter
```

The `ImportExternalFilterLoader` (which is associated with this line) will load `custom.filter` and try to call `SearchFilterFiles.getFilterByName( "positioning.filter" );`. This loads all filters again, which include `custom.filter` and then would load the `InternalFilterLoader`, ... -> infinite loop.

### Fix

The method `SearchFilterFiles#getFilterByName` will now only load the filters that match the name, resulting either in an empty or a single entry map. This is done by not directly loading the filters, i.e. keeping the `FilterLoader` and picking only the filters that match the name specified.

### State of this PR

I do not like this this approach and fix 100%, but it works for now.

### DIFF
--- a/src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
+++ b/src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
@@ -158,8 +158,8 @@ public class SearchFilterFiles {
 	}
 
 	public static Filter getFilterByName( final String name ) {
-		final Filter filter = toFileNameFilterMapping().get( name );
-		if ( filter == null ) {
+		final FilterLoader loader = toFileLoaderMapping().get( name );
+		if ( loader == null ) {
 			throw ProjectConfiguration.getInstance().getProjectConfigFolder() //
 					.map( path -> path.resolve( FILTER_DIR_NAME ) ) //
 					.map( Path::toAbsolutePath ) //
@@ -167,6 +167,6 @@ public class SearchFilterFiles {
 					.map( s -> new FilterNotFoundException( name, s ) )
 					.orElseGet( () -> new FilterNotFoundException( name ) );
 		}
-		return filter;
+		return loadSilently( loader, name );
 	}
 }
--- a/src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
+++ b/src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
@@ -17,7 +17,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
```diff
@@ -37,6 +37,23 @@
...
+						final FilterLoader loader = pair.getRight();
+						try {
+							return loader.load();
+						} catch ( final IOException e ) {
+							log.error( "Could not load filter for name '{}'.", pair.getLeft(), e );
+							return Filter.NEVER_MATCH;
+						}
+					},
+					// Prefer specific over generic filters (due to concat order).
+					( specificFilter, genericFilter ) -> specificFilter );
```
beatngu13: Since this is now a separate, class-wide collector, we don't control the stream order anymore. Before we had:

```java
Stream.of( getProjectFilterFiles(), getUserFilterFiles(), getDefaultFilterFiles() )
```

So the filter that came first was indeed more specific. But now one could apply the collector anywhere, like:

```java
Stream.of( getDefaultFilterFiles(), getUserFilterFiles(), getProjectFilterFiles() )
```

Then it would be `(  genericFilter, specificFilter ) -> genericFilter`.

Therefore, I suggest to rename the parameters and/or collector and add a few lines of Javadoc. Maybe name it `fifoCollector` and then have `( first, second ) -> first`?

src/test/java/de/retest/recheck/ignore/SearchFilterFilesTest.java
```diff
@@ -135,6 +136,25 @@ void getFilterByName_should_return_user_filter_with_given_name_over_default_filt
 		assertThat( result.toString() ).isEqualTo( "CompoundFilter(filters=[])" );
 	}
 
+	@Test
+	@ClearSystemProperty( key = RETEST_PROJECT_ROOT )
+	void getFilterByName_should_only_load_searched_filter() throws IOException {
+		System.setProperty( RETEST_PROJECT_ROOT, filterFolder.toString() );
+
+		// Create infinite filter loop which will cause infinite loop if loaded
+		final Path resolve = filterFolder.resolve( "foo.filter" );
```
beatngu13: Since you named below `bar`:

```suggestion
		final Path foo = filterFolder.resolve( "foo.filter" );
```

Although `bar` appears to be never used?

src/test/java/de/retest/recheck/ignore/SearchFilterFilesTest.java
```diff
@@ -135,6 +136,25 @@ void getFilterByName_should_return_user_filter_with_given_name_over_default_filt
...
+		Files.write( resolve, Collections.singleton( "import: bar.filter" ) );
+		final Path bar = filterFolder.resolve( "bar.filter" );
+		Files.write( resolve, Collections.singleton( "import: foo.filter" ) );
+
+		final Path positioningFilter = filterFolder.resolve( "content.filter" );
+		Files.createFile( positioningFilter );
+
+		final Filter result = SearchFilterFiles.getFilterByName( "content.filter" );
+		assertThat( result ).isNotNull();
+		assertThat( result.toString() ).isEqualTo( "CompoundFilter(filters=[])" );
```
beatngu13: Rather than asserting the string representation, AssertJ has:

```java
assertThat( result ).isInstanceOfSatisfying( CompoundFilter.class,
		compound -> assertThat( compound.getFilters() ).isEmpty() );
```

beatngu13 (changes requested):

src/test/java/de/retest/recheck/ignore/SearchFilterFilesTest.java
diba1013: That was copy-paste error. I fixed that.

src/main/java/de/retest/recheck/ignore/SearchFilterFiles.java
diba1013: That is a valid point. I pushed an alternative approach (I still have the old changes, if you wish).

Collect the `FilterLoader` in a map which does the loading strategy (i.e. fifo). Then load the filters or the specific filter, if present.

beatngu13 (approved):

PR #688, opened 2020-03-19
Update default filters with some more

diba1013:

- [x] ~the necessary tests are either created or updated.~ Don't think that is necessary
- [x] ~you updated the changelog.~ Don't think that is necessary
- [x]~ you updated necessary documentation within [retest/docs](https://github.com/retest/docs).~

## Description

> TL;DR Add some more filter rules.

When browsing through the differences for a twitter dashboard change, I noticed some attributes are not filtered. These are a couple of more rules to add.

### State of this PR

As we iteratively changing these filters, it could be definitely improved.

### DIFF
--- a/src/main/resources/filter/web/content.filter
+++ b/src/main/resources/filter/web/content.filter
@@ -1,6 +1,13 @@
 # Filter file for recheck that will filter content
 
 attribute=text
+attribute=title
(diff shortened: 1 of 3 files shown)


githubert (approved):
:+1:

PR #690, opened 2020-03-20
Well, tabindex is actually the correct attribute

diba1013:
Well, [`tabindex`](https://www.w3schools.com/tags/att_tabindex.asp) is actually the correct attribute. This is due to a [recent ](retest/recheck-web#561) change and should be migrated (i.e. accepted by the user) after the next release.


githubert (approved):
:+1: — That was the one attribute where I thought “I remember using that a while ago!”, well apparently not :D

PR #691, opened 2020-03-21
Feature/improve outline handling

roesslerj:

### DIFF
--- a/src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
+++ b/src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
@@ -53,7 +53,7 @@ public class AttributeUtil {
 		return outlineAttribute.getValue();
 	}
 
-	protected static Rectangle getActualOutline( final ElementDifference difference, final String type ) {
+	static Rectangle getActualOutline( final ElementDifference difference, final String type ) {
 		if ( !(difference.getIdentifyingAttributesDifference() instanceof IdentifyingAttributesDifference) ) {
 			log.error( "getIdentifyingAttributesDifference should return an IdentifyingAttributesDifference!" );
 			return difference.getElement().getIdentifyingAttributes().getOutlineRectangle();
--- a/src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
+++ b/src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
@@ -53,8 +53,8 @@ public class AttributeUtil {
 		return outlineAttribute.getValue();
 	}
 
-	private static Rectangle getActualOutline( final ElementDifference difference, final String type ) {
-		if (!(difference.getIdentifyingAttributesDifference() instanceof IdentifyingAttributesDifference)) {
+	protected static Rectangle getActualOutline( final ElementDifference difference, final String type ) {
+		if ( !(difference.getIdentifyingAttributesDifference() instanceof IdentifyingAttributesDifference) ) {
 			log.error( "getIdentifyingAttributesDifference should return an IdentifyingAttributesDifference!" );
 			return difference.getElement().getIdentifyingAttributes().getOutlineRectangle();
 		}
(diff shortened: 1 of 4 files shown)


src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
```diff
@@ -52,6 +54,10 @@ public static Rectangle getAbsoluteOutline( final IdentifyingAttributes attribut
 	}
 
 	private static Rectangle getActualOutline( final ElementDifference difference, final String type ) {
```
beatngu13: The method is called from two public methods, please add tests for the fixed behavior.

src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
```diff
@@ -52,6 +54,10 @@ public static Rectangle getAbsoluteOutline( final IdentifyingAttributes attribut
...
+		if (!(difference.getIdentifyingAttributesDifference() instanceof IdentifyingAttributesDifference)) {
```
beatngu13: `difference.getIdentifyingAttributesDifference()` is called twice, maybe extract the value earlier without cast?

beatngu13 (changes requested):
Please update the changelog with the bug fix.

src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
roesslerj: I think a method named `getIdentifyingAttributesDifference` not returning `IdentifyingAttributesDifference` is not intended behavior. Therefore I would not add a test to cement this further.

beatngu13: When you say this can / did cause bugs, there shall be tests. There is really no need to discuss this further.

roesslerj: Well. Adding a sensible test isn't easy at it involves downstream recheck projects. Adding just a mock test is cementing unwanted behavior.

roesslerj: So what do you suggest?

roesslerj: https://retest.atlassian.net/browse/RET-2015

beatngu13: It is recheck, why is there a need to involve anything else?

You can have two simple unit tests: Invoke both `getActualOutline` and `getActualAbsoluteOutline` with an `ElementDifference` where `getIdentifyingAttributesDifference()` does _not_ return an `IdentifyingAttributesDifference` but something else, then you get the `ClassCastException` you just fixed?

We have a lot of "unwanted behavior" like this. That doesn't mean there shouldn't be tests for it.

diba1013: > I think a method named getIdentifyingAttributesDifference not returning IdentifyingAttributesDifference is not intended behavior. Therefore I would not add a test to cement this further.

But isn't this fix then only a workaround? Wouldn't the actual fix fix the cause on why the method does not return the said class (because it either is `null` or actually a wrong instance)?

On the other hand, if this is wanted behavior and the method is just poorly named, I agree with @beatngu13 to add a test for that. At least to make sure that this is not "broken" while we work on the actual fix, which then can remove/modify the test added here.

src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
```diff
@@ -51,7 +53,11 @@ public static Rectangle getAbsoluteOutline( final IdentifyingAttributes attribut
 		return outlineAttribute.getValue();
 	}
 
-	private static Rectangle getActualOutline( final ElementDifference difference, final String type ) {
+	protected static Rectangle getActualOutline( final ElementDifference difference, final String type ) {
```
beatngu13: Can be package private:

```suggestion
	static Rectangle getActualOutline( final ElementDifference difference, final String type ) {
```

beatngu13 (changes requested):

src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
roesslerj: I think this is really just a getter, extracting this into a variable will bloat the code and make it less readable.

src/main/java/de/retest/recheck/ui/descriptors/AttributeUtil.java
roesslerj: An actual fix is probably way more complicated. I just added a test for that...

beatngu13 (approved):

PR #699, opened 2020-03-26
Min-width and height and vertical align are also positioning attributes

roesslerj:

Adding min-height, min-width and vertical-align to positioning.filter, as these govern element positions.

### DIFF
--- a/src/main/resources/filter/web/positioning.filter
+++ b/src/main/resources/filter/web/positioning.filter
@@ -7,6 +7,15 @@ attribute=bottom
 attribute=right
 attribute=left
 
+attribute=width
+attribute=height
+attribute=min-width
+attribute=min-height
+attribute=max-width
+attribute=max-height
+
+attribute=vertical-align
+
 attribute=clip
 
 attribute=margin-top


src/main/resources/filter/web/positioning.filter
```diff
@@ -7,6 +7,11 @@ attribute=bottom
 attribute=right
 attribute=left
 
+attribute=min-height
+attribute=min-width
```
diba1013: We can directly add all relevant attributes of https://developer.mozilla.org/en-US/docs/Web/CSS/min-width (see navigation), mainly:
```
width
height
max-width
max-height
```

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ Table of Contents
 
 ### Improvements
 
+* Adding min-height, min-width and vertical-align to positioning.filter, as these govern element positions.
```
diba1013: Thanks for updating the changelog. However, I recall that we agreed that for these changes we do not update the changelog.

diba1013 (changes requested):

diba1013 (approved):

PR #700, opened 2020-03-27
Run tests with newest JDK 14

beatngu13:


modulo11 (approved):

beatngu13:
Now fails for the same reason as https://github.com/retest/recheck-web/pull/572.

Will close both PRs until Sonar has sorted this out.

PR #719, opened 2020-04-14
Feature/ignore by attribute

roesslerj:


src/main/java/de/retest/recheck/review/ignore/matcher/ElementByAttributeMatcher.java
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	@Override
+	public boolean test( final Element element ) {
+		Object elementAttribute = element.getIdentifyingAttributes().get( key );
+		if ( elementAttribute != null ) {
+			return value.equals( elementAttribute );
+		}
+		elementAttribute = element.getAttributes().get( key );
+		return value.equals( elementAttribute );
```
lemoncurry, line 27: Just a suggestion to use different names for the local variables as well as Optional for clarity.
```suggestion
		final Optional<Object> elementIdentifyingAttribute =
				Optional.ofNullable( element.getIdentifyingAttributes().get( key ) );
		if ( elementIdentifyingAttribute.isPresent() ) {
			return value.equals( elementIdentifyingAttribute.get() );
		}
		final Object elementAttributes = element.getAttributes().get( key );
		return value.equals( elementAttributes );
```

src/test/java/de/retest/recheck/review/ignore/MatcherFilterLoaderTest.java
```diff
@@ -70,9 +70,9 @@ void load_should_produce_correct_ignore() {
 	}
 
 	@Test
-	void load_with_invalid_matcher_should_not_load_parent() throws Exception {
+	void load_with_generic_matcher_should_load_correctly() throws Exception {
```
lemoncurry:
```suggestion
	void load_with_generic_matcher_should_load_correctly() {
```

src/test/java/de/retest/recheck/review/ignore/MatcherFilterLoaderTest.java
```diff
@@ -89,8 +89,8 @@ void load_chained_should_produce_correct_ignore() {
 	}
 
 	@Test
-	void load_chained_with_invalid_matcher_should_not_load_parent() throws Exception {
-		final String line = "matcher: foo=bar, attribute: 123";
-		assertThat( cut.load( line ) ).isNotPresent();
+	void load_chained_with_generic_matcher_should_produce_correct_ignore() throws Exception {
```
lemoncurry:
```suggestion
	void load_chained_with_generic_matcher_should_produce_correct_ignore() {
```

src/test/java/de/retest/recheck/review/ignore/MatcherFilterLoaderTest.java
lemoncurry: `load_with_invalid_matcher_should_not_load_parent()` -> so there is no longer any invalid matcher possible with this new implementation?

lemoncurry (changes requested):

lemoncurry:
Additionally, [https://docs.retest.de/recheck/usage/filter/](https://docs.retest.de/recheck/usage/filter/) will need an update.

src/test/java/de/retest/recheck/review/ignore/MatcherFilterLoaderTest.java
roesslerj: Good point, I'll add a test

src/main/java/de/retest/recheck/review/ignore/matcher/ElementByAttributeMatcher.java
roesslerj: I feel that this makes the code harder to read...

lemoncurry: Hard to discuss 'feelings'.

Ideally `element.getIdentifyingAttributes().get( key )` should return an Optional<Object> (or even better, an Optional<someInterface>), not null or Object.
Personally, I prefer to use `isPresent()` instead of `!= null`. 

But of course this is your decision and this will not be taken into account for the PR approval.

src/test/java/de/retest/recheck/review/ignore/MatcherFilterLoaderTest.java
```diff
@@ -89,8 +89,15 @@ void load_chained_should_produce_correct_ignore() {
...
 	@Test
-	void load_chained_with_invalid_matcher_should_not_load_parent() throws Exception {
-		final String line = "matcher: foo=bar, attribute: 123";
+	void load_chained_with_generic_matcher_should_produce_correct_ignore() {
+		final String line = "matcher: foo=bar, attribute=123";
+		assertThat( cut.load( line ).map( cut::save ) ).hasValue( line );
+	}
+
+	@Test
+	void load_with_invalid_matcher_should_not_load_parent() throws Exception {
```
lemoncurry:
```suggestion
	void load_with_invalid_matcher_should_not_load_parent() {
```

lemoncurry (changes requested):

lemoncurry (approved):

PR #720, opened 2020-04-16
Attempt re-upload if report upload to rehub fails

lemoncurry:
- [ ] <s>the necessary tests are either created or updated.</s>

## Description

see https://github.com/retest/recheck/pull/720#issuecomment-623990874

### DIFF
--- a/src/main/java/de/retest/recheck/RecheckProperties.java
+++ b/src/main/java/de/retest/recheck/RecheckProperties.java
@@ -35,57 +35,57 @@ public interface RecheckProperties extends Reloadable {
 	 * Various constants.
 	 */
 
-	static final String PROPERTY_VALUE_SEPARATOR = ";";
-	static final String ZIP_FOLDER_SEPARATOR = "/";
+	String PROPERTY_VALUE_SEPARATOR = ";";
+	String ZIP_FOLDER_SEPARATOR = "/";
 
-	static final String SCREENSHOT_FOLDER_NAME = "screenshot";
-	static final String RECHECK_FOLDER_NAME = "recheck";
-	static final String DEFAULT_XML_FILE_NAME = "retest.xml";
+	String SCREENSHOT_FOLDER_NAME = "screenshot";
+	String RECHECK_FOLDER_NAME = "recheck";
+	String DEFAULT_XML_FILE_NAME = "retest.xml";
 
-	static final String RETEST_FOLDER_NAME = ".retest";
-	static final String RETEST_PROPERTIES_FILE_NAME = "retest.properties";
+	String RETEST_FOLDER_NAME = ".retest";
+	String RETEST_PROPERTIES_FILE_NAME = "retest.properties";
 
-	static final String GOLDEN_MASTER_FILE_EXTENSION = ".recheck";
-	static final String TEST_REPORT_FILE_EXTENSION = ".report";
-	static final String AGGREGATED_TEST_REPORT_FILE_NAME = "tests" + TEST_REPORT_FILE_EXTENSION;
+	String GOLDEN_MASTER_FILE_EXTENSION = ".recheck";
+	String TEST_REPORT_FILE_EXTENSION = ".report";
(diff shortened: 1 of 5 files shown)


diba1013:
Since @lemoncurry is currently busy, I have taken over this PR so that it can be included in the next release.

# Description

This PR aims to reduce errors that fail the build if the report upload fails. Because of unstable internet connections or large reports, this [fails quite often](https://github.com/diba1013/docs-test/actions?query=event%3Aschedule) although there are no real differences.

My assumptions are as follows:

1. If a report has no differences, it is fairly useless. Thus ignore any exceptions and do not attempt to upload, but log a visible warning message.
2. If a report has differences, retry `3` times (configurable via property `de.retest.recheck.rehub.upload.attempts`) and log a warning each time it fails. If max attempts have been reached, log an error and rethrow the exception.
3. The upload link (which theoretically can also throw exceptions, but it is ignored for now) is only generated once, because it is valid for 5min. If the upload fails within the timespan, we can assume it will not complete (we do not want to spam our services).

The steps taken are described below (please refer to the commits for a more detailed explanation):

1. Add property for max retry attempts on report upload.
2. Move report metadata to be only executed once (name, data), so that this is not re-evaluated on every retry.
3. Move reading of report to top, to have a better overview where the exceptions are thrown. This removes the need to catch all exceptions. 
   1. An `IOException` (file cannot be read) usually indicates a more severe problem, which cannot be fixed with a retry.
   2. The other methods only throw runtime exceptions, wrapped within a `UnirestException` (according a shallow dive into the code).

## State of this PR

Since it is fairly difficult to test this code (no tests have been written for the `CloudPersistence` yet), thus I omitted to write test for that. However, I will run several manual tests to verify that this works properly. In the meantime, this PR will stay a draft.

I will prepare a the documentation entry for this feature.

## Additional context

Please refer to RET-2021 for more information.

diba1013:
With my manual testing, this should work, although I only got an upload error a few times.

src/main/java/de/retest/recheck/RecheckProperties.java
```diff
@@ -35,57 +35,57 @@ static RecheckProperties getInstance() {
...
-	static final String RETEST_PROPERTIES_FILE_NAME = "retest.properties";
+	String RETEST_FOLDER_NAME = ".retest";
+	String RETEST_PROPERTIES_FILE_NAME = "retest.properties";
 
-	static final String GOLDEN_MASTER_FILE_EXTENSION = ".recheck";
-	static final String TEST_REPORT_FILE_EXTENSION = ".report";
-	static final String AGGREGATED_TEST_REPORT_FILE_NAME = "tests" + TEST_REPORT_FILE_EXTENSION;
+	String GOLDEN_MASTER_FILE_EXTENSION = ".recheck";
+	String TEST_REPORT_FILE_EXTENSION = ".report";
+	String AGGREGATED_TEST_REPORT_FILE_NAME = "tests" + TEST_REPORT_FILE_EXTENSION;
```
modulo11, line 50: Why are these (and the following) not final anymore?

src/main/java/de/retest/recheck/persistence/ReportUploadMetadata.java
```diff
@@ -1,15 +1,13 @@
...
 
-import java.net.URI;
-import java.util.List;
-
 import lombok.Builder;
 import lombok.Getter;
 
 @Builder
 @Getter
 public class ReportUploadMetadata {
```
modulo11, line 8: The name doesn't fit now. Suggestion: Strip the Metadata or rename to something like `ReportUploadContainer`.

src/main/java/de/retest/recheck/RecheckProperties.java
diba1013: Interface constants do not need these modifiers as they are already static final internally (`RecheckProperties.RECHECK_FOLDER_NAME = "bar"` will throw a compile error). I mostly followed the [Intellij Inspection "Unecessary interface modifier"](https://www.jetbrains.com/help/idea/list-of-java-inspections.html).

modulo11 (approved):

diba1013:
@Mergifyio refresh

PR #727, opened 2020-05-15
Release/v1.11.0

diba1013:


githubert (approved):
:+1: 

(Also while I suggested removing the `=3` from the property name in the documentation, because there is a list of properties with default values included, I think it is fine here in the context of release notes)

PR #732, opened 2020-06-03
Feature/metadata filtering

diba1013:

- [x] you updated necessary documentation within [retest/docs](https://github.com/retest/docs). (retest/docs#114)

## Description

> TL;DR Pass the recheck.ignore through to filter metadata

Instead of having a custom filter for the metadata differences, it now feeds of the global filter passed into `TestReportFilter`. 

However, this change now breaks the current behavior of filtering some volatile attributes. Thus I have added a new provided `metadata.filter` which can simply be overwritten and imported. Furthermore I added this to the `defaults-recheck.ignore` which gets applied to new projects.

To get this filter known to recheck (since there is a difference for the development [folder] and production environment [jar] and we must therefore statically define these filters), I have reworked the semi-fancy path loading to use the classical class resource loading. This fixes some rare cases of `java.nio.file.FileSystemAlreadyExistsException`.

### State of this PR

Although I have added the new metadata.filter to the `default-recheck.ignore` which gets placed within the project directory on initial test execution, it will not do so on subsequent executions (meaning the file already exists). However, I would not consider this a worthwhile entry as a breaking change, since the metadata differences do not have any effect on the reporting. Nonetheless, I will make sure, to properly communicate that in the upcoming release changelog and the documentation.

Secondly, I am currently passing a dummy/artificial element to emulate the `matches(Element, AttributeDifference)` filter and prevent NPEs for filters that attempt to access the data of said element. This is just filled with arbitrary data and might not conform to all edge cases. However, the actual metadata filters should never try to access the element (although in the future we might take that as an identifier, e.g: `matcher: type=metadata`). This will be properly documented.

### DIFF
--- a/src/main/resources/filter/metadata.filter
+++ b/src/main/resources/filter/metadata.filter
@@ -1,0 +1,9 @@
+attribute=machine.name
+attribute=os.arch
+attribute=vcs.branch
+attribute=vcs.commit
+attribute=vcs.name
+attribute=time.date
+attribute=time.time
+attribute=time.offset
+attribute=time.zone
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ Table of Contents
 
 ### New Features
 
-* Metadata differences are now filterable via the recheck.ignore.
+* Metadata differences are now filterable via the recheck.ignore. A new `metadata.filter` has been added which can be imported with `import: metadata.filter` in `recheck.ignore`.
 
 ### Improvements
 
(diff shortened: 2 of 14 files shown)


src/main/resources/filter/metadata.filter
```diff
@@ -0,0 +1,9 @@
+attribute=machine.name
```
roesslerj, line 1: Default is to filter all metadata? Maybe instead comment all line, but keep them so users know what metadata there is?

roesslerj (approved):

Issue #740, opened 2020-06-16
Cannot able to see the retest id in the log


ambily-padman:
when the UI changes happen i am expecting  to see  the retestId in the log, but it only shows the expected and actual as bellow:
_

```
> 4 check(s) in 'de.retest.quickstarts.LoginTest' found the following difference(s):
> Test 'login' has 6 difference(s) in 4 state(s):
> 00 resulted in:
> 	Metadata Differences:
> 	  Please note that these differences do not affect the result and are not included in the difference count.
> 		url:
> 		  expected="<url expected>",
> 		    actual="<url actual>"
input (username) at 'html[1]/body[1]/form[1]/div[1]/input[1]':
		name:
		  expected="username",
		    actual="user", breaks="LoginTest.java:84"
```

_

diba1013:
Hello @ambily-padman,

I apologize for taking so long to respond. I skimmed through the code and our [documentation](https://retest.de). The `retestId` is actually printed, however the documentation is not really clear and in some places inconsistent in that matter.

To clarify, the above report output consists of the following:

1. Structure representation
1. Metadata Differences
1. Element Differences
1. Attribute Differences

While the first two are solely to give some outlines and give hints on where/how the change occurred, the element differences are the important bit. It reads as follows:

```
type (retestid) at 'path':
    ... (some attribute differences)
```

1. `type`: The html tag name of the element (`<button>, <input>, ...`).
2. `retestid`: The stable virtual id (called `retestId`), which per [default](https://github.com/retest/recheck/blob/master/src/main/java/de/retest/recheck/ui/descriptors/idproviders/ElementCountingRetestIdProvider.java) looks for either `text`, `id`, `name` or `type` to be human readable. This can be [configured](https://docs.retest.de/recheck/usage/configuration/) with `RecheckOptionsBuilder#retestIdProvider(RetestIdProvider)`.
3. `path`: The recheck compatible XPath for that element.

Thus in your example, the retestId is the following `retestId=username`, which appears to be the name. Note that it will never change, despite the name changing in the displayed attribute difference.

I will try my best to update the documentation and necessary files to make this feature more clear and visible. Thank you for bringing this to our attention.

--- 

Since reading a wall of text is difficult, we try to keep the output as clear and concise as possible and therefore have decided against using something like 
`input (retestId=username) at '...'` at the expense that the user must familiarize themselves with the syntax.

However, we welcome any ideas, improvements and contributions to improve the readability of the report output.

[...]

PR #742, opened 2020-06-17
Add a changelog entry for the addition of a retest id

diba1013:

- [x] ~the necessary tests are either created or updated.~
- [x] you updated necessary documentation within [retest/docs](https://github.com/retest/docs). (*pending...*)

## Description

> TL;DR Apparently, we have missed to add a entry for displaying the `retestId`.

recheck-1.10.0 introduced the `retestId` within a report output. As #740 brought to my attention that we have not updated our documentation accordingly (and the changelog for that matter) to include the introduction of the retestId in the report output. Before (pre 1.10), it was a semi-random output of some basic attributes as #665 clarifies. Furthermore #667 ensured that the `retestId` was consistently displayed when a report is printed.

I have added the missing changelog entry and will update the documentation later.

### Additional Context

This change was introduced by a mix of #665 and #667 and thus we might have missed to add a respective changelog entry.


githubert (approved):
:+1:

PR #748, opened 2020-06-29
Fix/element warning

diba1013:

## Description

> TL;DR This fixes a problem with automatic code healing, where the breaking changes were lost when accepting attribute differences without these warnings.

### Background

The `GlobalChangeSetApplier` collects and aggregates related differences and accepts them in one go, based on a key created from both the element and difference. Once done, it will accept all differences with the difference passed, not the original difference (as this cannot be reconstructed).

There are multiple ways to settle this:

1. Do nothing, i.e. enforce the difference to contain a warning (aka the UI must pass the correct difference). This is difficult, as the UI itself uses caching to get notified for the related attributes.
1. Allow the original differences for each `ActionReplayResult` be reconstructable (i.e. restructurize the report).
1. Aggregate the collection of all warnings per element and difference and reconstruct the correct difference with warning.
1. ... *(more solutions I forgot or dismissed until now)*

I went for approach 3 because it is one of the easiest solutions, I came up with. Whenever an attribute is accepted/ignored (or the reverse actions), this will now attempt to reconstruct the difference with all encountered warnings. Since the `AttributeChanges` saves them simply within a list, all add/remove operations must be with the same object (a.k.a. the warnings).

### State of this PR

However, the `GlobalChangeSetApplier` now grows slowly and it becomes more apparent that the decision to add the warning to the `AttributeDifference` is not that great from an algorithmic point of view (both in `equals` and `apply`).

As mentioned in the commit message, this does not technically create the expected `ReviewResult`, because as soon as one accepted difference has an warning, all of the related differences will now contain warnings. These differences, however, are not present in the original report. But we never check for such a condition and thus I favored this approach over the others.

Furthermore, this essentially renders the passing of a `AttributeDifference` useless, as it will now only ever be used as a identifier. Thus we should rethink the collection of warnings in the future.

Lastly, this is fairly difficult to test, as a full integration test is a pain to setup. Thus I only added basic unit tests to (hopefully) ensure correct behavior in the future.

### Additional Context

This can be reproduced with the [auto-healing tutorial](https://docs.retest.de/recheck-web/usage/healing/). Notice that `btn-login` will not be healed with the broken version, since the initial encounter does not contain a warning added, and since this is cached, it will perform all operations with this difference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ Table of Contents
 
 ### Bug Fixes
 
+* Fix accepting of attributes without warning deletes warnings of similar attribute differences.
+
 ### New Features
 
 ### Improvements
--- a/src/main/java/de/retest/recheck/review/GlobalChangeSetApplier.java
+++ b/src/main/java/de/retest/recheck/review/GlobalChangeSetApplier.java
@@ -19,6 +19,7 @@ import de.retest.recheck.ui.diff.AttributeDifference;
 import de.retest.recheck.ui.diff.ElementDifference;
 import de.retest.recheck.ui.diff.InsertedDeletedElementDifference;
 import de.retest.recheck.ui.review.ActionChangeSet;
+import de.retest.recheck.ui.review.AttributeChanges;
 import lombok.AccessLevel;
 import lombok.Getter;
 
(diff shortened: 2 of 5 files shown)


githubert (approved):
As usual, a much appreciated thought-out description of the problem and potential solutions. :+1:

PR #750, opened 2020-06-30
Fix/attribute null

diba1013:

## Description

> TL;DR Fix NPE when attribute is inserted and applied

This is somewhat caused by #748 which destroys the attribute differences.

When performing the [autohealing tutorial](https://docs.retest.de/recheck-web/usage/healing/) applying the text attribute difference (which is an identifying attribute) results in an exception reported, because the attribute is missing and cannot be applied.

```text
java.lang.NullPointerException: Cannot apply change to an attribute that is null.
```

### Fix

The changes in #748 causes the attribute difference to be reconstructed. This destroys any special attribute differences (e.g. `AdditionalAttributeDifference`), which are used to convey additional meaning to the `IdentifyingAttributes` apply mechanism.

This was used to pass the original inserted attribute along and ignore the missing `null` attribute to simply create a new attribute (see `AdditionalAttributeDifference`). However, decided to custom handle the insertion of a (identifying) attribute the following:

1. If an attribute is present: Use the difference to apply.
2. If an attribute is missing: Check the actual value for attribute or try to create a new default attribute.

### State of this PR

I decided not to revert or adjust the changes made by the causing PR, since most custom `AttributeDifferences` are related to legacy code (to my understanding)  and should not be created. The remaining differences (mostly `AdditionalAttributeDifference`) should not require special handling (besides the one to be introduced).

### Additional Context

This bug has happened a few times in the past and couldn't be reproduced, since it seemed to happen only on edge cases. I hope, that this fix also fixes those cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Table of Contents
 ### Bug Fixes
 
 * Fix accepting of attributes without warning deletes warnings of similar attribute differences.
+* Fix NPE when applying an identifying attribute that has been inserted.
 
 ### New Features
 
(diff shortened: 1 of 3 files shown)


githubert (approved):
:+1:

PR #756, opened 2020-07-06
Add metadata for web, that should also be ignored on default...

roesslerj:
The name of the filter (metadata.filter) suggests that it filters all meta-data. So it should also filter recheck-web metadata. As far as I know, this hasn't made it to the docs yet, so...

### DIFF
--- a/src/main/resources/filter/metadata.filter
+++ b/src/main/resources/filter/metadata.filter
@@ -7,3 +7,11 @@ attribute=time.date
 attribute=time.time
 attribute=time.offset
 attribute=time.zone
+# for recheck-web
+attribute=browser.name
+attribute=browser.version
+attribute=check.type
+attribute=driver.type
+attribute=os.name
+attribute=os.version
+attribute=url
--- a/src/main/resources/filter/metadata.filter
+++ b/src/main/resources/filter/metadata.filter
@@ -1,12 +1,5 @@
-attribute=machine.name
-attribute=os.arch
-attribute=vcs.branch
-attribute=vcs.commit
-attribute=vcs.name
-attribute=time.date
-attribute=time.time
-attribute=time.offset
-attribute=time.zone
+import: volatile-metadata.filter
+
 # for recheck-web
 attribute=browser.name
 attribute=browser.version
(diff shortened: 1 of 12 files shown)


diba1013 (changes requested):
The intention behind the basic implementation was to establish sensible defaults, not to exclude everything. Particularly, it should ignore ever-changing attributes (time, ...), but keep important attributes, since they may be the cause for the changes (i.e. browser name changed).

To use other words, the implementation was focused to keep the current implementation (1.11.0) which still reports the important attributes. I would urge that the user specify any additional attributes themselves.

I understand the confusion and wanted to keep the name as simple as possible (for the general, use cases) but I think that I would prefer to have a separate filter file for all metadata e.g. `metadata-all.filter`.

roesslerj:
Hmm. How about `default-metadata.filter` and `metadata.filter`, where the first is default-included in the `recheck.ignore` file and the latter ignores all metadata?

diba1013:
> Hmm. How about default-metadata.filter and metadata.filter, where the first is default-included in the recheck.ignore file and the latter ignores all metadata?

Something along this line, maybe emphasize the volatile-ness of the attributes (e.g. `metadata-volatile.filter`). I prefer to have the metadata as first to directly specify the type (see bootstrap `btn-primary`), but I will leave that up to you. Be sure to update the documentation as outlined in retest/docs#114 and the `default-recheck.ignore`.

src/main/resources/filter/metadata.filter
```diff
@@ -7,3 +7,11 @@ attribute=time.date
 attribute=time.time
 attribute=time.offset
 attribute=time.zone
```
diba1013: Now you could use `import: volatile-metadata.filter` here too 😄

src/test/java/de/retest/recheck/RecheckImplIT.java
```diff
@@ -30,7 +30,7 @@
 	static final String MESSAGE_START_PATTERN = "A detailed report will be created at '.*'";
 	static final String MESSAGE_START_REPLACE = "A detailed report will be created at '*'";
 
-	static final Filter METADATA_FILTER = Filters.parse( "import: metadata.filter" );
+	static final Filter METADATA_FILTER = Filters.parse( "import: volatile-metadata.filter" );
```
diba1013: As seen in #757, I changed that to exclude all metadata, so this can left as `metadata.filter` (optional).

diba1013 (changes requested):
I forgot to mention that you should update the changelog entry as well...

diba1013 (approved):

PR #759, opened 2020-07-07
Feature/reduce redundancy

roesslerj:

## Description

TL;DR: Just remove some redundancy by using the same code...


diba1013 (approved):
I think that I need to take a look at this at some later time, because the initial intent was that the loaders should always return an optional and never throw an exception, so that other loaders have the opportunity to take over the loading mechanism. Now you throw an exception in every case, which ultimately makes the `MatcherFilterLoader` the only loader who is able to load `matcher: ...`.

roesslerj:
Hmm. Ok, I can see that. But anyway, this behavior is then the same for every situation...

PR #761, opened 2020-07-07
Log _why_ an imported filter cannot be loaded...

roesslerj:
- [ ] all status checks (Travis CI, SonarCloud, etc.) pass.

## Description

TL;DR: Log _why_ an imported filter cannot be loaded...

### DIFF
--- a/src/main/java/de/retest/recheck/review/ignore/io/ImportExternalFilterLoader.java
+++ b/src/main/java/de/retest/recheck/review/ignore/io/ImportExternalFilterLoader.java
@@ -27,7 +27,7 @@ public class ImportExternalFilterLoader extends RegexLoader<ImportedExternalFilt
 			final Filter loaded = SearchFilterFiles.getFilterByName( reference );
 			return Optional.of( new ImportedExternalFilter( reference, loaded ) );
 		} catch ( final Exception e ) {
-			log.error( "Exception loading referenced filter '{}'.", reference );
+			log.error( "Exception loading referenced filter '{}': {}.", reference, e.getMessage() );
 			return Optional.empty();
 		} catch ( final StackOverflowError e ) {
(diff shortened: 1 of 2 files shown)


src/main/java/de/retest/recheck/review/ignore/io/ImportExternalFilterLoader.java
```diff
@@ -27,7 +27,7 @@ public ImportExternalFilterLoader() {
 			final Filter loaded = SearchFilterFiles.getFilterByName( reference );
 			return Optional.of( new ImportedExternalFilter( reference, loaded ) );
 		} catch ( final Exception e ) {
-			log.error( "Exception loading referenced filter '{}'.", reference );
+			log.error( "Exception loading referenced filter '{}': {}", reference, e );
```
diba1013: Exceptions do not need a parameter. It suffices to do the following:
```suggestion
			log.error( "Exception loading referenced filter '{}'.", reference, e );
```
Otherwise, use only the message
```suggestion
			log.error( "Exception loading referenced filter '{}': {}.", reference, e.getMessage() );
```

diba1013 (changes requested):

src/main/java/de/retest/recheck/review/ignore/io/ImportExternalFilterLoader.java
roesslerj: Hmm. I tried both. Problem is that your first suggestion logs the complete stack trace (much too verbose), while the second option ... well unfortunately, `FileNotFoundException` doesn't have a message...

diba1013: And what does your implementation output?

diba1013 (approved):

PR #763, opened 2020-07-10
Feature/detect project layout

diba1013:

- [x] you updated necessary documentation within [retest/docs](https://github.com/retest/docs). (*retest/docs#121*)

## Description

> TL;DR Instead of always using Maven, this performs checks to detect to see if the respective `pom.xml` or `build.gradle` as a central file.

The change is fairly simple:

1. Make the `ProjectLayout` optional instead of having a default.
2. If not defined by the time `RecheckOptions#build` is called, perform some checks to select the appropriate layout.

The detection mechanism only utilizes simple checks, which should suffice for the majority of setups (as explained above). In the future, we might need to do some more advanced checks (similar to `PathBasedProjectRootFinder`) and have multiple indicators. In such a case, we could take a look at what IDEs do when opening such folders.

Note that this will only work for simple use cases. Things like multiple directories with different build systems, multiple build systems or custom configurations of the build systems cannot be reflected with the auto detect. However, in this case I would argue that customizations must be reflected through a respective custom `ProjectLayout`, as we currently use the *convention-over-configuration* and not the *zero-configuration* policy.

Because of simplicity and flaws related to the project discovery (see below), the method layout detection does not differentiate between "project not found" and "no pom.xml or build.gradle file present"

### Initialization

In order to ensure that recheck is properly initialized, I have moved the initialization phase into `RecheckOptionsBuilder#build`. This is related to minimize flaws with the project discovery (see below).

### Testing

All tests are performed within a temporary directory to emulate and configure a clean project. Secondly, they each configure the base folder for the project root discovery (instead of using the execution directory) within `setUp`. But because of the way the project discovery works, especially tests that assume an empty project directory, may fail because some parent directory (e.g. user home) defines a `.retest` folder and is discovered through walking up the file system structure.

Furthermore, most methods `RecheckOptionsBuilderIT` create a project, because the builder might need it (e.g. filter files) for additional checks. Such checks are present in `ProjectLayoutsIT`.

### State of this PR

I am unsure, if we should use the project directory as discovered by `ProjectRootFinderUtil` or use the execution directory as `Paths#get`. The former allows for easier testing as a temporary directory can be used and utilizes the common approach for project discovery. However, as described by the `RecheckOptionsIT`, there are flaws in the project discovery. If there is no project and a custom project layout (i.e. no `src/main/java`) is used, the project directory cannot be identified, and thus an exception is thrown. Therefore we might need to rethink the initial project setup in such a case (i.e. `ProjectConfiguration#ensureProjectConfigurationInitialized`). The workaround for now is to create a `.retest` folder manually.

Although, this changes the default behavior, which could be seen as a breaking change, I would not consider it as such, because the `ProjectLayout` is such an critical component that has to be set. The only use case where this would be a breaking change, if the user never configured it (thus uses maven) and just accepted that the reports are saved within `target` (ignoring it in git or similar). Thus, with the update, this use case would either correctly save reports within `build` (in case of Gradle) or throw (in case of unsupported). If you think different, we could always add an additional warning inside *breaking changes*. In either case, I will double check and make sure that the updated documentation specifies the importance of the project layout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Table of Contents
 * Metadata differences are now filterable via the recheck.ignore. New `volatile-metadata.filter` and `metadata.filter` filters have been added, where the first one filters metadata which is added for diagnosing purposes and too verbose to be displayed upon every change and the latter filters _all_ metadata. They can be imported with `import: volatile-metadata.filter` and `import: metadata.filter` in `recheck.ignore`.
 * Add `exclude()` filter expression to allow child elements to be excluded. Ignoring an element `matcher: type=body, exclude(matcher: type=button)` will ignore everything below the `body` element but still capture changes for `button` elements.
 * Add a `disableReportUpload` method.
+* The `ProjectLayout` will now be automatically detected (Maven and Gradle) and will throw an error if it cannot fall back to Maven, in which case a custom `ProjectLayout` has to be provided.
 
 ### Improvements
 
--- a/src/main/java/de/retest/recheck/RecheckOptions.java
+++ b/src/main/java/de/retest/recheck/RecheckOptions.java
@@ -18,6 +18,7 @@ import de.retest.recheck.persistence.GradleProjectLayout;
 import de.retest.recheck.persistence.MavenProjectLayout;
 import de.retest.recheck.persistence.NamingStrategy;
 import de.retest.recheck.persistence.ProjectLayout;
+import de.retest.recheck.persistence.ProjectLayouts;
 import de.retest.recheck.ui.descriptors.idproviders.RetestIdProvider;
 import de.retest.recheck.util.RetestIdProviderUtil;
 import lombok.AccessLevel;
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ Table of Contents
 * Metadata differences are now filterable via the recheck.ignore. New `volatile-metadata.filter` and `metadata.filter` filters have been added, where the first one filters metadata which is added for diagnosing purposes and too verbose to be displayed upon every change and the latter filters _all_ metadata. They can be imported with `import: volatile-metadata.filter` and `import: metadata.filter` in `recheck.ignore`.
 * Add `exclude()` filter expression to allow child elements to be excluded. Ignoring an element `matcher: type=body, exclude(matcher: type=button)` will ignore everything below the `body` element but still capture changes for `button` elements.
 * Add a `disableReportUpload` method.
+* The `ProjectLayout` will now be automatically detected (Maven and Gradle) and will throw an error if it cannot instead of falling back to Maven, in which case a custom `ProjectLayout` has to be provided.
```
githubert:
```suggestion
* The `ProjectLayout` will now be automatically detected (Maven and Gradle) and will throw an error if it cannot fall back to Maven, in which case a custom `ProjectLayout` has to be provided.
```

githubert (changes requested):
:+1: — Looks nice, just having a minor suggestion for the changelog.

githubert (approved):
:+1:

PR #799, opened 2020-10-14
Feature/fix pixel diff filter npe

martin-v:
Prevent NPE in PixelDiffFilter


roesslerj (commented):
Is it possible to have a test for that?

PR #997, opened 2022-04-08
Fix AttributeUtilTest

martin-v:


dkrn03 (commented):
OK, but why is resetTestInstance() only called in some tests? What is the indicator for when to use it?

martin-v:
`resetTestInstance()` should be used every time a test instance ist created with `getTestInstance()`.

dkrn03 (approved):

